PR #45, opened 2017-11-29
TSK-15: Create a manual task in a given Workbasket

BVier:


BVier:
I'm not all sure about the npm build.
At first it worked just for admin, after installing it again on monitor it worked for both. 
So maybe it was just a local problem, but it would be good to check this on another system.

PR #52, opened 2017-12-07
Tsk 61

MLengl:
imtoduced classificationinterface
improved unittests

### DIFF
--- a/lib/taskana-cdi/src/test/java/pro/taskana/TaskanaRestTest.java
+++ b/lib/taskana-cdi/src/test/java/pro/taskana/TaskanaRestTest.java
@@ -1,18 +1,24 @@
 package pro.taskana;
 
+import javax.ejb.EJB;
+import javax.ws.rs.DELETE;
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.core.Response;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import pro.taskana.exceptions.ClassificationAlreadyExistException;
 import pro.taskana.exceptions.ClassificationNotFoundException;
 import pro.taskana.exceptions.NotAuthorizedException;
 import pro.taskana.exceptions.TaskNotFoundException;
 import pro.taskana.exceptions.WorkbasketNotFoundException;
-import pro.taskana.model.Classification;
+import pro.taskana.model.ClassificationImpl;
 import pro.taskana.model.Task;
 import pro.taskana.model.Workbasket;
-
-import javax.ejb.EJB;
-import javax.ws.rs.*;
-import javax.ws.rs.core.Response;
 
 @Path("/test")
 public class TaskanaRestTest {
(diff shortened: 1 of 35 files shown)


lib/taskana-cdi/src/test/java/pro/taskana/TaskanaRestTest.java
```diff
@@ -23,12 +29,12 @@
...
 
 	@GET
-	public Response startTask() throws NotAuthorizedException, WorkbasketNotFoundException, ClassificationNotFoundException {
+	public Response startTask() throws NotAuthorizedException, WorkbasketNotFoundException, ClassificationNotFoundException, ClassificationAlreadyExistException {
 		Workbasket workbasket = new Workbasket();
 		workbasket.setName("wb");
 		taskanaEjb.getWorkbasketService().createWorkbasket(workbasket);
-		Classification classification = new Classification();
-		taskanaEjb.getClassificationService().addClassification(classification);
+		Classification classification = (Classification) new ClassificationImpl();
```
BerndBreier: We should behave like a client in this situation. I.e. don't use ClassificationImpl but create a classification via ClassificationService.newClassification().

lib/taskana-core/src/main/java/pro/taskana/exceptions/ClassificationNotFoundException.java
```diff
@@ -1,6 +1,7 @@
 package pro.taskana.exceptions;
 
 /**
+<<<<<<< HEAD
  * This exception will be thrown if a specific task is not in the database.
  */
```
BerndBreier: Javadoc should say: ..thrown if a specific Classification is not in the database

lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationServiceImpl.java
```diff
@@ -72,17 +75,30 @@ public ClassificationServiceImpl(TaskanaEngine taskanaEngine, ClassificationMapp
...
+        try {
             taskanaEngineImpl.openConnection();
-            classification.setId(IdGenerator.generateWithPrefix(ID_PREFIX_CLASSIFICATION));
-            classification.setCreated(Date.valueOf(LocalDate.now()));
-
-            this.setDefaultValues(classification);
-
-            classificationMapper.insert(classification);
-            LOGGER.debug("Method addClassification added classification {}.", classification);
+            ClassificationImpl classificationImpl = (ClassificationImpl) classification;
```
BerndBreier: we obtain a Classification from the client. Before we continue, we should ensure that it is a valid object. This implies for instance that the Id is set ...

MLengl: setDefaultValues() is called before the insert-call. setting an ID if is/was null is included there.

lib/taskana-cdi/src/test/java/pro/taskana/TaskanaRestTest.java
MLengl: I will search and fix all classification-casts which are like this. The instantiation by service-call will be added in the upcoming pull-request.

PR #61, opened 2017-12-12
TSK-22_start_working_on_a_task

BerndBreier:


lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
```diff
@@ -54,27 +59,48 @@ public TaskServiceImpl(TaskanaEngine taskanaEngine, TaskMapper taskMapper,
         this.taskanaEngineImpl = (TaskanaEngineImpl) taskanaEngine;
         this.taskMapper = taskMapper;
         this.objectReferenceMapper = objectReferenceMapper;
+        this.workbasketService =  taskanaEngineImpl.getWorkbasketService();
     }
 
     @Override
-    public Task claim(String id, String userName) throws TaskNotFoundException {
+    public Task claim(String id)  throws TaskNotFoundException, InvalidStateException, InvalidOwnerException {
```
MLengl: Why do we need two different claim() Methods? And maybe it should be called claimTask().

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
```diff
@@ -54,27 +59,48 @@ public TaskServiceImpl(TaskanaEngine taskanaEngine, TaskMapper taskMapper,
...
+        return claim(id, false);
+    }
+
+    @Override
+    public Task claim(String id, boolean forceClaim)  throws TaskNotFoundException, InvalidStateException, InvalidOwnerException {
+        String userName = CurrentUserContext.getUserid();
+        return claim(id, userName, forceClaim);
+    }
+
+    public Task claim(String id, String userName, boolean forceClaim) throws TaskNotFoundException, InvalidStateException, InvalidOwnerException {
```
MLengl: should be private and only get invoked by _this.claim()_
UserName is not necessary as parameter, because CurrentUser.getUserName() is static.

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
```diff
@@ -54,27 +59,48 @@ public TaskServiceImpl(TaskanaEngine taskanaEngine, TaskMapper taskMapper,
...
         LOGGER.debug("entry to claim(id = {}, userName = {})", id, userName);
         Task task = null;
         try {
             taskanaEngineImpl.openConnection();
             task = taskMapper.findById(id);
```
MLengl: Should call this.getTaskById(), because DRY-principle gets hurt, if we repeat exceptionhandling/reactions on other functionalities.

PR #63, opened 2017-12-12
TSK-23: Complete a Task with force-flag and user-check. 

MLengl:
Added Flag to completeTask().
Added owner check to completeTask
Added unit test for this function.
Changed RELEASE-Info/Dependency to 0.0.2
Updated JavaDoc

- IntegrationTests should be done after full CLAIM and Authentification integration.

### DIFF
--- a/lib/taskana-core/src/main/java/pro/taskana/TaskService.java
+++ b/lib/taskana-core/src/main/java/pro/taskana/TaskService.java
@@ -1,7 +1,11 @@
 package pro.taskana;
 
+import java.util.List;
+
 import pro.taskana.exceptions.ClassificationNotFoundException;
 import pro.taskana.exceptions.NotAuthorizedException;
+import pro.taskana.exceptions.NotOwnerException;
+import pro.taskana.exceptions.TaskNotClaimedException;
 import pro.taskana.exceptions.TaskNotFoundException;
 import pro.taskana.exceptions.WorkbasketNotFoundException;
 import pro.taskana.model.DueWorkbasketCounter;
@@ -9,8 +13,6 @@ import pro.taskana.model.Task;
 import pro.taskana.model.TaskState;
 import pro.taskana.model.TaskStateCounter;
 import pro.taskana.model.TaskSummary;
-
-import java.util.List;
 
 /**
  * The Task Service manages all operations on tasks.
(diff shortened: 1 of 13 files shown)


lib/taskana-core/src/main/java/pro/taskana/TaskService.java
```diff
@@ -29,13 +31,18 @@
...
+     * @param isForced - Flag which can complete a Task in every case if Task does exist.
+     *
+     * @return Task - updated task after completion.
+     *
+     * @throws TaskNotClaimedException when Task wasn´t claimed before.
+     * @throws TaskNotFoundException if the given Task can´t be found in DB.
+     * @throws NotOwnerException if current user is not the task-owner or administrator.
      */
-    Task complete(String taskId) throws TaskNotFoundException;
+    Task completeTask(String taskId, boolean isForced) throws TaskNotClaimedException, TaskNotFoundException, NotOwnerException;
```
MLengl: Should be seperated into **_completeTask(String taskId)_** and **_completeTask(String taskId, boolean isForced)_**

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
```diff
@@ -83,26 +86,36 @@ public Task claim(String id, String userName) throws TaskNotFoundException {
     }
 
     @Override
-    public Task complete(String id) throws TaskNotFoundException {
-        LOGGER.debug("entry to complete(id = {})", id);
+    public Task completeTask(String taskId, boolean isForced) throws TaskNotClaimedException, TaskNotFoundException, NotOwnerException {
```
MLengl: Should be seperated into **_completeTask(String taskId)_** and **_completeTask(String taskId, boolean isForced)_**

**_completeTask(String taskId)_** just needs to call **_completeTask( , )_** with false

lib/taskana-core/src/test/java/pro/taskana/impl/TaskServiceImplTest.java
```diff
@@ -27,6 +27,7 @@
 import org.mockito.Mock;
```
MLengl: Need to add _**completeTask(String taskId)**_ test to prove it uses false-flag

MLengl: Added verification that **isForced = false** is given to _completeTask( , );_

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
MLengl: Done.

lib/taskana-core/src/main/java/pro/taskana/TaskService.java
MLengl: Done

MLengl:
Waited for TSK-22 and added the wanted changes.
Rebased on current HEAD
--> new Pullrequest incoming.

PR #66, opened 2017-12-13
TSK-39: Created a Interface for TASK-Entity

MLengl:
Created Task-Interface
Replaced all public available Services with this Interface
Unit/Integrationtest - Fixed


MLengl:
Waiting to solve conflicts and make unit-test running again.

PR #71, opened 2017-12-14
TSK-23: Complete a Task... (REOPEN)

MLengl:
...with force-flag and user-check. 
Added unit test for this function. 
Added completeTask(id) without flag
improved unit-test
reacted to TSK-22.


MLengl:
new squashed update of older pull-request: [#63](https://github.com/Taskana/taskana/pull/63)

PR #91, opened 2018-01-08
TSK-136: acceptance test for getAttachments

holgerhagen:


MLengl:
There are some errors while dropping DB-Tables (Attachment) and Test is not Running correctly (CreateTaskAccTest).  Please update.

PR #99, opened 2018-01-11
TSK-146: Introduced ClassificationSummary Interface and Impl.

MLengl:
Also deleted forbidden methods on Classification-Interface. 
Mappers for Query changed.

### DIFF
--- a/lib/taskana-core/src/main/java/pro/taskana/ClassificationQuery.java
+++ b/lib/taskana-core/src/main/java/pro/taskana/ClassificationQuery.java
@@ -5,18 +5,20 @@ import java.util.Date;
 /**
  * ClassificationQuery for generating dynamic sql.
  */
-public interface ClassificationQuery extends BaseQuery<Classification> {
+public interface ClassificationQuery extends BaseQuery<ClassificationSummary> {
 
     /**
      * Add your key to your query.
+     *
      * @param key
-     *           as String
+     *            as String
      * @return the query
      */
     ClassificationQuery key(String... key);
 
     /**
      * Add your parentClassificationKey to your query.
+     *
      * @param parentClassificationKey
      *            as String
      * @return the query
@@ -25,6 +27,7 @@ public interface ClassificationQuery extends BaseQuery<Classification> {
 
     /**
      * Add your category to your query.
+     *
      * @param category
      *            as String
      * @return the query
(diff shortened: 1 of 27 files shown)


lib/taskana-core/src/main/java/pro/taskana/ClassificationQuery.java
```diff
@@ -41,36 +45,44 @@
 
     /**
      * Add your domain to your query.
-     * @param domain TODO
+     *
+     * @param domain
+     *            TODO
```
BerndBreier: Please fill in javadoc 'TODO' if modifying a class / interface

lib/taskana-core/src/main/java/pro/taskana/ClassificationQuery.java
```diff
@@ -41,36 +45,44 @@
...
      * @return TODO
      */
     ClassificationQuery domain(String... domain);
 
     /**
      * Add to your query if the Classification shall be valid in its domain.
-     * @param validInDomain TODO
+     *
+     * @param validInDomain
+     *            TODO
```
BerndBreier: see above

lib/taskana-core/src/main/java/pro/taskana/ClassificationQuery.java
```diff
@@ -41,36 +45,44 @@
...
      * @return TODO
      */
     ClassificationQuery validInDomain(Boolean validInDomain);
 
     /**
      * Add your created-Dates to your query.
-     * @param created TODO
+     *
+     * @param created
+     *            TODO
```
BerndBreier: also here

lib/taskana-core/src/main/java/pro/taskana/ClassificationSummary.java
```diff
@@ -0,0 +1,59 @@
+package pro.taskana;
```
BerndBreier: Is it really required to have interface + implementation for summaries?
TaskSummary and WorkbasketSummary are simple Java Beans in the pro.taskana.model folder.
What is the reason for I/F + Impl for ClassificationSummary?

MLengl: All model-classes we´ve got implements an interface for the user-api. So all summaries should implement an interface, too - i´ve asked Holger before i´ve done it this way.

MLengl:
Commit is updated and complete.

PR #100, opened 2018-01-12
TSK-147 Introduce Workbasket summary to be returned in lists and from…

BerndBreier:
… all queries.


BerndBreier:
should have been updated when I rebased and pushed my branch to BerndBreier/taskana

PR #108, opened 2018-01-16
TSK-139 update a classification

BerndBreier:


lib/taskana-core/src/test/java/acceptance/classification/UpdateClassificationAccTest.java
```diff
@@ -44,20 +47,51 @@ public void testUpdateClassification() throws SQLException, ClassificationNotFou
         classification.setName("newName");
         classification.setParentClassificationKey("T2000");
         classification.setPriority(1000);
-        classification.setServiceLevel("newServiceLevel");
+        try {
+            classification.setServiceLevel("newServiceLevel");
+        } catch (IllegalArgumentException ignored) {
+
+        }
         // setType and setDomain must be made impossible! (remove setter from interface)
```
MLengl: Setters has been removed so this comments should be removed, too.

lib/taskana-core/src/test/java/pro/taskana/impl/integration/ClassificationServiceImplIntExplicitTest.java
```diff
@@ -123,17 +122,16 @@ public void testInsertClassification()
 
         // does exist already
         try {
-            expectedClassification = (ClassificationImpl) this.createNewClassificationWithUniqueKey();
+            expectedClassification = (ClassificationImpl) this.createNewClassificationWithUniqueKey(domain, "t1");
             expectedClassification.setKey(key);
```
MLengl: This line is reduncant, becasue Key+Domain are given line above.

BerndBreier: Not really. createNewClassificationWithUniqueKey creates -as the name implies- a classification with a unique key. The purpose of the test is to create a ClassificationAlreadyExistsException. To do this, it overrides the unique key that was created in method createNewClassificationWithUniqueKey by a duplicate key thats stored in variable key.

PR #109, opened 2018-01-16
TSK-149 Use embedded primary ObjectReference without helper fields

BerndBreier:


lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
```diff
@@ -444,31 +440,15 @@ public Attachment newAttachment() {
...
-                objRef.setCompany(attachment.getPorCompany());
-                objRef.setSystem(attachment.getPorSystem());
-                objRef.setSystemInstance(attachment.getPorSystemInstance());
-                objRef.setType(attachment.getPorType());
-                objRef.setValue(attachment.getPorValue());
-                attachment.setObjectReference(objRef);
-                results.add(attachment);
+    private void setAttachments(TaskImpl result) {
+        List<AttachmentImpl> attachmentImpls = attachmentMapper.findAttachmentsByTaskId(result.getId());
+        List<Attachment> attachments = new ArrayList<Attachment>();
```
MLengl: <> can/should be used empty

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
```diff
@@ -444,31 +440,15 @@ public Attachment newAttachment() {
...
+        if (attachmentImpls != null && !attachmentImpls.isEmpty()) {
+            for (AttachmentImpl attImpl : attachmentImpls) {
```
MLengl: can be written as 1line lambda. (optinal)

BerndBreier: As we saw, lambda performance is worse than the simple loop performance

PR #111, opened 2018-01-17
TSK145-2: Add Acc Test for task transfer

BVier:


lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationServiceImpl.java
```diff
@@ -323,8 +323,11 @@ public Classification getClassificationByTask(Task task) throws ClassificationNo
             result = classificationMapper.findByTask(classificationKey, task.getWorkbasketKey(),
                 CURRENT_CLASSIFICATIONS_VALID_UNTIL);
             if (result == null) {
+                result = classificationMapper.findByKeyAndDomain(classificationKey, "", CURRENT_CLASSIFICATIONS_VALID_UNTIL);
```
MLengl: Waht´s the difference between byKeyAndDomain() and byTask()? The Query is using a JOIN and we should avoid this. GetByKeyAndDomain() should be finde. Just a single call. 
ClassificationMapper.findByTask can be deleted and shoulöd be deleted.

lib/taskana-core/src/test/java/acceptance/task/TransferTaskAccTest.java
```diff
@@ -0,0 +1,92 @@
...
+    @WithAccessId(
+            userName = "user_1_1",
+            groupNames = { "group_1" })
+    @Test
+    public void testDomainChangingWhenTransferTask()
+            throws SQLException, NotAuthorizedException, InvalidArgumentException, ClassificationNotFoundException,
+            WorkbasketNotFoundException, TaskAlreadyExistException, InvalidWorkbasketException, TaskNotFoundException {
+
+        TaskService taskService = taskanaEngine.getTaskService();
+
```
MLengl: Please write codes in blocks in this calss.
Code is simple and can be written as 1-2 blocks (by functionallity seperated) - Have a look the other lines in this class.

lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationServiceImpl.java
holgerhagen: this change should be removed from the PR. getClassificationByTask should be deleted. But this is another story.

PR #113, opened 2018-01-18
TSK-83: delete Classification (with tests)

BVier:
Please check the assumption behind "careForTheChildren".

I wasn't sure if I should make one or two functions out of it. 
The two-function-case seemed saver to me.

Last but not least: we could realize "getClassificationOfTask" with less database-queries. (we could use the task-domain instead of the workbasket-domain) 
but this is maybe not that save... Just want to say it!


lib/taskana-core/src/main/java/pro/taskana/ClassificationService.java
```diff
@@ -44,6 +46,34 @@
...
+     * @param domains
+     *              the domains for which you want to delete the classification.
+     * @throws ClassificationInUseException
+     *              if there are Task existing, which refer to this classification.
+     * @throws NotAuthorizedException
+     *              if you're not authorized to delete the classification.
+     * @throws ClassificationNotFoundException
+     *              if for an domain no classification specification is found.
+     */
+    void deleteClassification(String classificationKey, String... domains) throws ClassificationInUseException, NotAuthorizedException, ClassificationNotFoundException;
```
MLengl: NotAuthorizedException shouldn´t be thrown here. It´s a Exception for Workbasket-Permissions and not for changing classifications. 
I guess you can ignore this case, because who ever will be able to see a data in a workbasket with a classification got at least READ permission.

**Please ask @holgerhagen if there is a special permission for this interaction or if you can remove this throwing argument.**

lib/taskana-core/src/main/java/pro/taskana/ClassificationService.java
```diff
@@ -44,6 +46,34 @@
...
+     * Delete a root-classification with all children.
+     *
+     * @param classificationKey
+     *              the key of the classification you want to delete.
+     * @throws ClassificationInUseException
+     *              if there are Task existing, which refer to this classification.
+     * @throws NotAuthorizedException
+     *              if you're not authorized to delete the classification.
+     */
+    void deleteRootClassification(String classificationKey) throws ClassificationInUseException, NotAuthorizedException;
```
MLengl: Have a look at the comment above. (line 63)

lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationServiceImpl.java
```diff
@@ -351,4 +352,78 @@ public Classification getClassificationByTask(Task task) throws ClassificationNo
         }
     }
 
+    @Override
+    public void deleteClassification(String classificationKey, String... domains) throws ClassificationInUseException, NotAuthorizedException, ClassificationNotFoundException {
+        if (domains.length == 0) {
```
MLengl: This can throw a NullPointer, because String.../String[] can be given as NULL. Please change to:
_( ==null || length==0)_

lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationServiceImpl.java
```diff
@@ -351,4 +352,78 @@ public Classification getClassificationByTask(Task task) throws ClassificationNo
...
+            throw new IllegalArgumentException("You have to specify a domain");
+        }
+        for (String domain : domains) {
+            if (domain.equals("")) {
+                // Alternative Solution: open "deleteRootClassification".
+                // But in my opinion there is a big difference between deleting a specification of a Classification and deleting the whole classification.
+                throw new IllegalArgumentException("You cannot delete the Root Classification this way. Please use 'deleteRootClassification' instead");
```
MLengl: Why not? Please check if there are Classifications which git the current one as PARENT and if there are no results, you can delete it here, too. 
It´s a special case, sry.

lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationServiceImpl.java
```diff
@@ -351,4 +352,78 @@ public Classification getClassificationByTask(Task task) throws ClassificationNo
...
+            }
+        }
+
+        TaskServiceImpl taskService = (TaskServiceImpl) taskanaEngineImpl.getTaskService();
+        boolean existingTasksForClassificationInDomains = ((TaskServiceImpl) taskService).isThereATaskForThisClassificationKeyInTheseDomains(classificationKey, domains);
```
MLengl: Other Entities/Models/... got Classifications, too. Sry, but you need to change the name of the variable and also the query to test it. 
Just COUNT references and give back a number >0 = there are references and you can´t delete. This would be readable.

lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationServiceImpl.java
```diff
@@ -351,4 +352,78 @@ public Classification getClassificationByTask(Task task) throws ClassificationNo
...
+        if (existingTasksForClassificationInDomains) {
+            throw new ClassificationInUseException("There are Tasks which belong to this classification. Please complete them and try again.");
+        }
+
+        this.internalDeleteClassification(classificationKey, domains);
```
MLengl: Here´s no try-catch but in the other method? Please make it constantly with one call-pattern.

lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationServiceImpl.java
```diff
@@ -351,4 +352,78 @@ public Classification getClassificationByTask(Task task) throws ClassificationNo
...
+            classification.setValidUntil(Date.valueOf(LocalDate.now().minusDays(1)));
+            classificationMapper.update(classification);
+        }
+    }
+
+    /**
+     * This function is based on the assumption, that a domain classification never has another parent than its root classification.
+     * In this case the children of a deleted domain classification fall automatically back to the root classification.
+     */
+    private void careForTheChildren(String classificationKey, String newParent) throws NotAuthorizedException {
```
MLengl: There are to many call´s and bad readability. Please use caliic or longer names for lists/variables. Why are you removing a single root/parent-classification like this?

lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationServiceImpl.java
```diff
@@ -351,4 +352,78 @@ public Classification getClassificationByTask(Task task) throws ClassificationNo
...
+        String newParent = classificationMapper.findByKeyAndDomain(classificationKey, "", CURRENT_CLASSIFICATIONS_VALID_UNTIL).getParentClassificationKey();
+
+        try {
+            this.internalDeleteClassification(classificationKey, domains);
+        } catch (ClassificationNotFoundException e) {
+            // should not happen.
+            throw new InternalError("A internal error occured. There may be a problem with the 'getDomainsForClassificationKey' request of the classificationMapper.");
+        }
+
+        this.careForTheChildren(classificationKey, newParent);
```
MLengl: Is a calling this function in every case (Exception) really necessary? Correct me if i´m wrong.

lib/taskana-core/src/main/java/pro/taskana/impl/TaskQueryImpl.java
```diff
@@ -37,7 +37,7 @@
     private String[] classificationKey;
     private String[] workbasketKey;
     private String[] domain;
-    private String[] owner;
+    private String[] owner2;
```
MLengl: Was temporal rename finding an error. Please change it back to owner.^^

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
```diff
@@ -316,6 +316,16 @@ public TaskQuery createTaskQuery() {
         return (results == null) ? new ArrayList<>() : results;
     }
 
+    public boolean isThereATaskForThisClassificationKeyInTheseDomains(String classificationKey, String... domain) throws NotAuthorizedException {
```
MLengl: See other comments. I guess this need to be updated with naming and parameter checks.

lib/taskana-core/src/main/java/pro/taskana/ClassificationService.java
holgerhagen: Delete classification will need a role based authorization. But we don't have it right know. So Marcel is right. NotAuthorized should not happen.

lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationServiceImpl.java
```diff
@@ -351,4 +352,78 @@ public Classification getClassificationByTask(Task task) throws ClassificationNo
...
+        this.internalDeleteClassification(classificationKey, domains);
+    }
+
+    @Override
+    public void deleteRootClassification(String classificationKey) throws ClassificationInUseException, NotAuthorizedException {
+        String[] domains = classificationMapper.getDomainsForClassificationKey(classificationKey, CURRENT_CLASSIFICATIONS_VALID_UNTIL);
+
+        TaskServiceImpl taskService = (TaskServiceImpl) taskanaEngineImpl.getTaskService();
+        List<Task> tasks = new ArrayList<>();
+        tasks.addAll(taskService.createTaskQuery().classificationKeyIn(classificationKey).list());
```
holgerhagen: A simple check on the size of the results would be enough, wouldn't it?

lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationServiceImpl.java
holgerhagen: Please move this to a method in ClassificationService and just use a TaskQuery. This method could throw the exception right away...

lib/taskana-core/src/main/java/pro/taskana/ClassificationService.java
holgerhagen: It is not the root classification. It is the classification from the "master domain" (domain == ""). Maybe we don't need a seperate method?

lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationServiceImpl.java
holgerhagen: Why don't we use a recursive function here?

lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationServiceImpl.java
holgerhagen: Actually we better delete all children as well instead of manipulating the tree.

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
holgerhagen: better don't even create a method in TaskService for deleting classification. The logic belongs to ClassificationService.

BVier:
Sry for the conflicting files... :/

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
```diff
@@ -835,9 +835,9 @@ private void handleAttachmentsOnTaskUpdate(TaskImpl oldTaskImpl, TaskImpl newTas
         }
     }
 
-    private void initAttachment(AttachmentImpl attachment, Task newTask) {
-        if (attachment.getId() == null) {
-            attachment.setId(IdGenerator.generateWithPrefix(ID_PREFIX_ATTACHMENT));
+    public int countTasksByClassificationAndDomain(String classificationKey, String domain) {
```
holgerhagen: Warum nicht einfach über eine TaskQuery? Dann bräuchte man keine eigene Methode...

holgerhagen: und man könnte die count-Funktion nutzen, die es jetzt bzw. bald gibt.

lib/taskana-core/src/test/java/acceptance/classification/DeleteClassificationAccTest.java
```diff
@@ -24,54 +22,85 @@
...
-        // classificationService.deleteClassification("L1050", "");
+    public void testDeleteMasterClassification()
+        throws SQLException, ClassificationNotFoundException, NotAuthorizedException, ClassificationInUseException {
+        classificationService.deleteClassification("L12010", "");
 
-        Classification classification = classificationService.getClassification("L140101", "DOMAIN_A");
-        assertTrue(classification == null);
+        boolean classificationNotFound = false;
+        try {
+            classificationService.getClassification("L12010", "DOMAIN_A");
```
holgerhagen: warum arbeitest du hier nicht mit expected = ...?

BVier: weil ich da noch den rollback teste

PR #116, opened 2018-01-22
Tsk 140

MiguelMartinRojas:


MiguelMartinRojas:
handlebars.js v 1.3.3 was causing the vulnerability security issue, now everything is updated. And warning should dissapear from github.

PR #118, opened 2018-01-24
TSK-125: Create/Update/Enable Acc-Tests.

MLengl:
especially for querying WB content.

### DIFF
--- a/lib/taskana-core/src/test/java/acceptance/task/UpdateTaskAccTest.java
+++ b/lib/taskana-core/src/test/java/acceptance/task/UpdateTaskAccTest.java
@@ -1,17 +1,23 @@
 package acceptance.task;
 
+import static org.hamcrest.core.IsEqual.equalTo;
+import static org.hamcrest.core.IsNot.not;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThat;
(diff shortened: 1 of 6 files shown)


lib/taskana-core/src/test/java/acceptance/task/UpdateTaskAccTest.java
```diff
@@ -136,64 +141,57 @@ public void testThrowsExceptionIfTaskHasAlreadyBeenUpdated()
...
-        // Task updatedTask = taskService.updateTask(task);
-        //
-        // assertNotNull(updatedTask);
-        // meaningful assertions
+        throws TaskNotFoundException, WorkbasketNotFoundException, ClassificationNotFoundException,
+        InvalidArgumentException, ConcurrencyException, InvalidWorkbasketException, NotAuthorizedException {
+        TaskService taskService = taskanaEngine.getTaskService();
+        Task task = taskService.getTask("TKI:000000000000000000000000000000000000");
+        task.setCustom1("T2100");
+        Task updatedTask = taskService.updateTask(task);
```
holgerhagen: Please get the task again for assertion of the test

MLengl: DONE.
Updated this for all update() Results and added fail() on testThrowsExceptionIfMandatoryPrimaryObjectReferenceIsNotSetOrIncomplete(), because there wouldn´t be a failure when there is now exception thrown in the tested cases.

PR #120, opened 2018-01-24
TSK-72: Read workbasket metadata

BVier:
The "getPermissionsForWorkbasketKey" method is really ugly, but I didn't find a better way to query this.
If there's a better solution, please tell it to me. (or just implement it)

The testing was really fast. Is there something missing?


BVier:
I reduced the query to one summarized workbasketAccessItem, but I still need the dirty code.
A case-Statement didn't work for me.

PR #126, opened 2018-01-29
TSK-155: Extend workbasket level report by subdivision into cluster

kklaeger:
- Create method that creates the report by a list of workbaskets, a list
of states and a list of cluster
- Create methods for initialization, create sumline and add counts in
the ReportLines
- Exclude tasks with DUE = null from the query
- Add method for manipulating SQL data
- Set databaseId to allow database specific queries in the mapper
- Add sql sample data for monitoring
- Create an own monitor test data method in TestDataGenerator
- ProvideWorkbasketLevelReportAccTest doesn't extend AbstractAccTest
anymore
- Remove integration tests and replace them by an acceptance test
- Spring files are not changed. They accidentally appear in this pull request


lib/taskana-core/src/main/java/pro/taskana/model/mappings/TaskMonitorMapper.java
```diff
@@ -52,14 +52,21 @@ long getTaskCountForWorkbasketByDaysInPastAndState(@Param("workbasketId") String
         @Param("status") List<TaskState> states);
 
     @Select("<script>"
-        + "SELECT WORKBASKET_KEY, COUNT(WORKBASKET_KEY) as counter "
+        + "<if test=\"_databaseId == 'db2'\">SELECT WORKBASKET_KEY, (DAYS(DUE) - DAYS(CURRENT_TIMESTAMP)) as AGE_IN_DAYS, COUNT(*) as NUMBER_OF_TASKS</if> "
+        + "<if test=\"_databaseId == 'h2'\">SELECT WORKBASKET_KEY, DATEDIFF('DAY', CURRENT_TIMESTAMP, DUE) as AGE_IN_DAYS, COUNT(*) as NUMBER_OF_TASKS</if> "
```
MLengl: What happens, if the current __databaseId_ is eighter "db2" or "h2"?
I´m guessing a _SqlException_ will be thrown - how do we want to deal with it?

_GROUPBY_ does work for each case, so it can be ignored as optional part.

lib/taskana-core/src/test/java/pro/taskana/database/TestDataGenerator.java
```diff
@@ -1,69 +1,127 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DbSchemaCreator.class);
+
+    private StringWriter outWriter = new StringWriter();
+
+    private PrintWriter logWriter;
+
+    private StringWriter errorWriter;
+
+    private PrintWriter errorLogWriter;
```
MLengl: Can you please remove the spaces between the attributes? They aren´t annotated so there should be no space. THX

PR #149, opened 2018-02-06
TSK-244: Reset TaskState after a transfer.

MLengl:


MLengl:
Fix missed import from TSK-251  - Ignore on TransferTaskAccTest

PR #156, opened 2018-02-06
TSK-48 Sort the results of a task query

BerndBreier:


lib/taskana-core/src/main/java/pro/taskana/impl/TaskQueryImpl.java
```diff
@@ -384,11 +395,21 @@ public String getPorValueLike() {
...
     @Override
     public String toString() {
         StringBuilder builder = new StringBuilder();
         builder.append("TaskQueryImpl [taskanaEngineImpl=");
         builder.append(taskanaEngineImpl);
+        builder.append(", taskService=");
+        builder.append(taskService);
         builder.append(", name=");
         builder.append(Arrays.toString(name));
         builder.append(", description=");
```
MLengl: Why is description include here, but not enabled in the mapper? Sorting description is useless but field is not selected/set currently. (QueryMapper L.64)

PR #161, opened 2018-02-07
TSK-252: Bulk-Transfer for Task UPDATED

MLengl:
Tested with a Acc-Test.
Method is declared at TaskService, not all Services!

Unittest will be implemented later, because I just want to get some feedback before.

### DIFF
--- a/lib/taskana-core/src/main/java/pro/taskana/impl/BulkOperationResults.java
+++ b/lib/taskana-core/src/main/java/pro/taskana/impl/BulkOperationResults.java
@@ -1,0 +1,90 @@
+package pro.taskana.impl;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Returning type for a bulk db interaction with errors. This wrapper is storing them with a matching object ID.
+ *
+ * @param <K>
+ *            unique keys for the logs.
+ * @param <V>
+ *            type of the stored informations
+ */
+public class BulkOperationResults<K, V> {
+
+    private Optional<Map<K, V>> errorMap = Optional.of(new HashMap<K, V>());
+    private static final Logger LOGGER = LoggerFactory.getLogger(BulkOperationResults.class);
+
+    /**
+     * Returning a list of current errors as map. If there are no errors the result will be empty.
+     *
+     * @return map of errors which can´t be null.
+     */
+    public Map<K, V> getErrorMap() {
+        return this.errorMap.get();
+    }
+
+    /**
+     * Adding an appearing error to the map and list them by a unique ID as key. NULL keys will be ignored.
+     *
+     * @param objectId
+     *            unique key of a entity.
+     * @param error
+     *            occurred error of a interaction with the entity
+     * @return status of adding the values.
+     */
+    public boolean addError(K objectId, V error) {
+        boolean status = false;
+        try {
+            if (objectId != null) {
+                this.errorMap.get().put(objectId, error);
+                status = true;
+            }
+        } catch (Exception e) {
+            LOGGER.warn(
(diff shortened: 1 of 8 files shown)


lib/taskana-core/src/main/java/pro/taskana/impl/BulkOperationResults.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Returning the status of a bulk-error-log.
+     *
+     * @return true if there are logged errors.
+     */
+    public boolean containErrors() {
+        boolean isContainingErrors = false;
+        if (this.errorMap.isPresent()) {
+            if (this.errorMap.get().size() > 0) {
```
BerndBreier: It is recommended to us isEmpty() method of collections to check for emptyness (rather than size() > 0 )

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
```diff
@@ -278,6 +280,89 @@ public Task transfer(String taskId, String destinationWorkbasketKey)
...
+                    .filter(taskSummary -> currentTaskId.equals(taskSummary.getTaskId()))
+                    .findFirst()
+                    .isPresent()) {
+                    bulkLog.addError(currentTaskId, new TaskNotFoundException(currentTaskId));
+                    taskIdIterator.remove();
+                }
+            }
+
+            // check source WB (read)+transfer
+            List<String> workbasketKeys = new ArrayList<>();
```
BerndBreier: would be better to create a Set of workbasketkeys, since a set contains each element only once

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
```diff
@@ -278,6 +280,89 @@ public Task transfer(String taskId, String destinationWorkbasketKey)
...
+                .keyIn(workbasketKeys.toArray(new String[0]))
+                .list();
+            taskIdIterator = taskIds.iterator();
+            while (taskIdIterator.hasNext()) {
+                String currentTaskId = taskIdIterator.next();
+                TaskSummary taskSummary = taskSummaries.stream()
+                    .filter(t -> currentTaskId.equals(t.getTaskId()))
+                    .findFirst()
+                    .get();
+                if (!sourceWorkbaskets.stream()
```
BerndBreier: findFirst() returns an Optional. .get() should only be called if you are sure that the optional is not empty. Otherwise you may want to call .orElse(null) rather than .get()

lib/taskana-core/src/main/java/pro/taskana/impl/BulkOperationResults.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Returning a list of current errors as map. If there are no errors the result will be empty.
+     *
+     * @return map of errors which can´t be null.
+     */
+    public Map<K, V> getErrorMap() {
+        if (!this.errorMap.isPresent()) {
+            this.errorMap = Optional.of(new HashMap<K, V>());
+        }
```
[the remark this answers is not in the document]
MLengl: done

lib/taskana-core/src/main/java/pro/taskana/impl/BulkOperationResults.java
MLengl: done

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
MLengl: done

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
MLengl: done

PR #163, opened 2018-02-07
TSK-250: CancelClaim with force-flag.

MLengl:


lib/taskana-core/src/main/java/pro/taskana/TaskService.java
```diff
@@ -56,6 +56,38 @@ Task claim(String taskId)
...
+     */
+    Task cancelClaim(String taskId) throws TaskNotFoundException, InvalidStateException, InvalidOwnerException;
+
+    /**
+     * Unclaim a existing Task which was claimed and owned by you before. Also there can be enabled a force flag for
+     * admins.
+     *
+     * @param taskId
+     *            id of the task which should be unclaimed.
+     * @return updated unclaimed task
```
BerndBreier: Javadoc for forceUnclaim missing. In addition, the name forceCancel or simply force for this parameter would be better

PR #173, opened 2018-02-12
Tsk-157: services return summaries and remove task-getter with wb-keys

BVier:
The services (workbasket, classification and task) now return summaries, except for get- and new- queries.
Further two queries in the TaskService were replaced by taskQueries (getTaskByWorkbasketIdAndState and getTaskSummariesByWorkbasketKey)

### DIFF
--- a/lib/taskana-core/src/test/java/pro/taskana/impl/WorkbasketServiceImplTest.java
+++ b/lib/taskana-core/src/test/java/pro/taskana/impl/WorkbasketServiceImplTest.java
@@ -31,6 +31,7 @@ import pro.taskana.TaskQuery;
 import pro.taskana.TaskService;
 import pro.taskana.TaskSummary;
 import pro.taskana.Workbasket;
+import pro.taskana.WorkbasketSummary;
 import pro.taskana.configuration.TaskanaEngineConfiguration;
 import pro.taskana.exceptions.InvalidArgumentException;
 import pro.taskana.exceptions.InvalidWorkbasketException;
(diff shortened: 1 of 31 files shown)


lib/taskana-core/src/test/java/pro/taskana/impl/WorkbasketServiceImplTest.java
```diff
@@ -335,8 +336,17 @@ public void testCreateWorkbasket_WithoutDistibutionTargets()
         verifyNoMoreInteractions(taskQueryMock, taskServiceMock, workbasketMapperMock, workbasketAccessMapperMock,
             distributionTargetMapperMock,
             taskanaEngineImplMock, taskanaEngineConfigurationMock);
-        assertThat(actualWb.getId(), not(equalTo(null)));
-        assertThat(actualWb.getId(), startsWith("WBI"));
+        assertThat(actualWbSummary.getId(), not(equalTo(null)));
+        assertThat(actualWbSummary.getId(), startsWith("WBI"));
+
+        doNothing().when(cutSpy).checkAuthorization(any(), any());
```
MLengl: A unit-test should only test 1 call per method. ALso there are few unused lines which should be removed or tested, too.

lib/taskana-core/src/test/java/pro/taskana/impl/WorkbasketServiceImplTest.java
```diff
@@ -370,15 +380,17 @@ public void testCreateWorkbasket_WithDistibutionTargets()
             taskanaEngineImplMock, taskanaEngineConfigurationMock);
         assertThat(actualWb.getId(), not(equalTo(null)));
         assertThat(actualWb.getId(), startsWith("WBI"));
-        assertThat(actualWb.getCreated(), not(equalTo(null)));
-        assertThat(actualWb.getModified(), not(equalTo(null)));
+//        assertThat(actualWb.getCreated(), not(equalTo(null)));
```
MLengl: Why "//"?  remove these lines every where if they´re not needed, but correct logic if they should behaviour like this.

PR #175, opened 2018-02-13
TSK-104 make constructor of impl classes package private

BerndBreier:


BerndBreier:
Constructor of TaskanaEngineImpl only be made protected because SpringTaskanaEngineImpl inherits from TaskanaEngineImpl and needs access to the constructor.
In addition, TaskanaEngineConfiguration needs to create a TaskanaEngine. For this purpose, I introduced the factory method TaskanaEngineImpl.createTaskanaEngine() ...

PR #184, opened 2018-02-15
TSK-276 Classification.CREATE in the schema should be a Timestamp rat…

BerndBreier:
…her than a Date


lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationQueryImpl.java
```diff
@@ -587,13 +467,13 @@ public String toString() {
...
-        builder.append(", created=");
-        builder.append(Arrays.toString(created));
+        builder.append(", createdIn=");
+        builder.append(Arrays.toString(createdIn));
         builder.append(", nameIn=");
         builder.append(Arrays.toString(nameIn));
         builder.append(", nameLike=");
         builder.append(Arrays.toString(nameLike));
-        builder.append(", descriptionLike=");
+        builder.append(", description=");
```
MLengl: There is "DescriptionLike" as filter at the QueryMapper. So the naming was correct i guess.

lib/taskana-core/src/main/java/pro/taskana/model/mappings/QueryMapper.java
```diff
@@ -136,7 +136,7 @@
         + "<if test='type != null'>AND TYPE IN(<foreach item='item' collection='type' separator=',' >#{item}</foreach>)</if> "
         + "<if test='domain != null'>AND DOMAIN IN(<foreach item='item' collection='domain' separator=',' >#{item}</foreach>)</if> "
         + "<if test='validInDomain != null'>AND VALID_IN_DOMAIN = #{validInDomain}</if> "
-        + "<if test='created != null'>AND CREATED IN(<foreach item='item' collection='created' separator=',' >SUBSTRING(#{item}, 1, 10)</foreach>)</if> "
+        + "<if test='createdIn !=null'> AND ( <foreach item='item' collection='createdIn' separator=' OR ' > ( <if test='item.begin!=null'> CREATED &gt;= #{item.begin} </if> <if test='item.begin!=null and item.end!=null'> AND </if><if test='item.end!=null'> CREATED &lt;=#{item.end} </if>)</foreach>)</if> "
         + "<if test='nameIn != null'>AND NAME IN(<foreach item='item' collection='nameIn' separator=',' >#{item}</foreach>)</if> "
         + "<if test='nameLike != null'>AND (<foreach item='item' collection='nameLike' separator=' OR '>NAME LIKE #{item}</foreach>)</if> "
         + "<if test='descriptionLike != null'>AND DESCRIPTION like #{descriptionLike}</if> "
```
MLengl: s.o.

lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationQueryImpl.java
BerndBreier: ok, I renamed the 'description' variable, although this has no effect on the correct method name .

lib/taskana-core/src/main/java/pro/taskana/model/mappings/QueryMapper.java
BerndBreier: createdIn is the name of an internal variable and doesn't need to confirm to naming conventions for operations.

PR #219, opened 2018-02-22
TSK-286: Replace workbaskets bei workbasketIds in TaskMonitorService

kklaeger:
- Set workbasket id /classification id in the report instead of the key
- Update acceptance and unit tests
- Remove unnecessary mocks in the unit test


lib/taskana-core/src/test/java/acceptance/monitoring/ProvideDetailedClassificationReportAccTest.java
```diff
@@ -68,50 +65,58 @@ public static void resetDb() throws SQLException, IOException {
...
     @Test
-    public void testGetTotalNumbersOfTasksOfClassificationReport()
+    public void testGetTotalNumbersOfTasksOfDetailedClassificationReport()
         throws WorkbasketNotFoundException, NotAuthorizedException, ClassificationNotFoundException {
 
         TaskMonitorService taskMonitorService = taskanaEngine.getTaskMonitorService();
 
-        List<Workbasket> workbaskets = getListOfWorkbaskets();
+        List<String> workbasketIds = Arrays.asList(
+            "WBI:000000000000000000000000000000000001",
```
MLengl: Optonal: You´re generating Lists/Single Objects with ID multiple time on few test-methods. Maybe you can create a method wich does generate X-IDs, starting at Y.          Array[]/List<Ids> ids = generateIds(3, 1);

Only in cases where you got only on ID you need to use get(0)/[0] or keep them like they are. Maybe method will very usefull if tests get´s lager.

PR #220, opened 2018-02-22
TSK-325 replaced taskana-core with taskana-spring to gain transaction…

mustaphazorgati:
… management

### DIFF
--- a/rest/pom.xml
+++ b/rest/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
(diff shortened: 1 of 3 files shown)


rest/pom.xml
```diff
@@ -33,10 +33,18 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-security</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-jdbc</artifactId>
```
tituslabs: isn't this include in spring-boot-start-jdbc?

rest/src/main/java/pro/taskana/rest/RestApplication.java
```diff
@@ -31,6 +37,8 @@
 import pro.taskana.sampledata.SampleDataGenerator;
 
 @SpringBootApplication
+@PropertySource("classpath:datasource.properties")
```
tituslabs: may be we change to the default application.properties. we already use the prefix datasource for this.

rest/src/main/java/pro/taskana/rest/RestApplication.java
```diff
@@ -40,23 +48,40 @@ public static void main(String[] args) {
...
+        return new DataSourceProperties();
     }
 
     @Bean
-    public TaskService getTaskService() throws Exception {
-        return getTaskanaEngine().getTaskService();
+    @Primary
+    public DataSource dataSource(DataSourceProperties properties) {
+        DataSource dataSource = properties.initializeDataSourceBuilder().build();
+        // if TaskanaEngineImpl runs with SpringManagedTransactionFactory, then
```
tituslabs: you can remove this comment or make it more clear to use TransactionAwareDataSourceProxy if using native jdbc api..

PR #231, opened 2018-02-26
TSK-316: Add unchangeable CREATOR to Task.

MLengl:


lib/taskana-core/src/main/java/pro/taskana/Task.java
```diff
@@ -19,6 +19,13 @@
      */
     String getId();
 
+    /**
+     * Gets the name of the task-creator.
```
holgerhagen: please replace name with userid

lib/taskana-core/src/main/resources/sql/taskana-schema.sql
```diff
@@ -63,6 +62,7 @@ CREATE TABLE TASK (
         PLANNED TIMESTAMP NULL,
         DUE TIMESTAMP NULL,
         NAME VARCHAR(255) NULL,
+        CREATOR VARCHAR(255) NULL,
```
holgerhagen: please change to VARCHAR(32). That should be fine for just userid (no group names!)

PR #238, opened 2018-02-26
Tsk329: Method removed and read permission checked in WorkbasketQuery

holgerhagen:


lib/taskana-core/src/main/java/pro/taskana/mappings/QueryMapper.java
```diff
@@ -231,9 +231,10 @@
         + "<if test='orgLevel3Like != null'>AND (<foreach item='item' collection='orgLevel3Like' separator=' OR ' >UPPER(w.ORG_LEVEL_3) LIKE #{item}</foreach>)</if> "
         + "<if test='orgLevel4In != null'>AND UPPER(w.ORG_LEVEL_4) IN(<foreach item='item' collection='orgLevel4In' separator=',' >#{item}</foreach>)</if> "
         + "<if test='orgLevel4Like != null'>AND (<foreach item='item' collection='orgLevel4Like' separator=' OR ' >UPPER(w.ORG_LEVEL_4) LIKE #{item}</foreach>)</if> "
-        + "<if test='authorization != null'>AND "
+        + "AND (PERM_READ = 1 "
+        + "<if test='authorization != null'>OR "
```
MLengl: I guessed count() just needs the READ permission and with an OR as link other permissions can override missing READ - is it realy the way we want this call?

lib/taskana-core/src/main/java/pro/taskana/mappings/QueryMapper.java
```diff
@@ -451,9 +451,10 @@
         + "<if test='orgLevel3Like != null'>AND (<foreach item='item' collection='orgLevel3Like' separator=' OR ' >UPPER(w.ORG_LEVEL_3) LIKE #{item}</foreach>)</if> "
         + "<if test='orgLevel4In != null'>AND UPPER(w.ORG_LEVEL_4) IN(<foreach item='item' collection='orgLevel4In' separator=',' >#{item}</foreach>)</if> "
         + "<if test='orgLevel4Like != null'>AND (<foreach item='item' collection='orgLevel4Like' separator=' OR ' >UPPER(w.ORG_LEVEL_4) LIKE #{item}</foreach>)</if> "
-        + "<if test='authorization != null'>AND "
+        + "AND (PERM_READ = 1 "
```
MLengl: same here

PR #249, opened 2018-02-28
TSK-341: No Exception when Classification-parent for ID does not exist

MLengl:
Changed ClassificationService.create() and update():
--> Throwing an Exception when no matching classification for the ParentClassificationId can be found.

update()
- removed insert, when classificationToUpdate does not ExistAlready (throwing notFoundException)
- removed getClassification in the end because it´s getting the same current classification again. Tests should approve that mapper.update() does work.

### DIFF
--- a/lib/taskana-core/src/main/java/pro/taskana/ClassificationService.java
+++ b/lib/taskana-core/src/main/java/pro/taskana/ClassificationService.java
@@ -65,20 +65,22 @@ public interface ClassificationService {
      *             when the classification does already exists at the given domain.
      * @throws NotAuthorizedException
      *             if the current user is not member of role BUSINESS_ADMIN or ADMIN
+     * @throws ClassificationNotFoundException
+     *             if the current parentId is not NULL/EMPTY and can not be found.
      */
     Classification createClassification(Classification classification)
-        throws ClassificationAlreadyExistException, NotAuthorizedException;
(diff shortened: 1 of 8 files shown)


lib/taskana-core/src/main/java/pro/taskana/ClassificationService.java
```diff
@@ -75,9 +77,9 @@ Classification createClassification(Classification classification)
      *            the Classification to update
      * @return the updated Classification.
      * @throws ClassificationNotFoundException
-     *             when the classification does not exist already.
+     *             when the classification OR it´s parent does not exist.
      * @throws NotAuthorizedException
-     *             when a user got no permissions for WB content tasks.
+     *             when a user got no permissions for WB content tasks if the category has changed.
```
holgerhagen: wird nur geworfen, wenn der Caller nicht BUSINESS_ADMIN oder ADMIN ist, oder?

lib/taskana-core/src/main/java/pro/taskana/impl/ClassificationServiceImpl.java
```diff
@@ -118,44 +121,30 @@ public Classification updateClassification(Classification classification) throws
             this.initDefaultClassificationValues(classificationImpl);
 
             // UPDATE/INSERT classification
```
holgerhagen: nur noch UPDATE?

PR #254, opened 2018-03-01
TSK-69: Offer all values of a DB-Column with filter by Query.

MLengl:
listValues() gets a TABLENAME as parameter and got the same QRY-String (with filters/...) like other QRYs of it´s type.
The returning value is a List of Strings containing all values which are used in the given column for the matching entries.

currently there is no parameter-validation! Maybe sould be a switch-case for columnnames later with InvalidArgumentException.
All Types can be converted to String (bool = 0/1, date="mm:dd:yyy..", ...)


MLengl:
@holgerhagen PR is up to date with sorting - Exception for not existing column for query will be PersistanceEx or need to be added.

PR #279, opened 2018-03-14
TSK-353 Extend the number of available custom attributes on the task

BerndBreier:


lib/taskana-core/src/main/java/pro/taskana/TaskQuery.java
```diff
@@ -403,204 +405,36 @@
     TaskQuery customFieldsIn(String... customFields);
```
holgerhagen: was ist das?

PR #283, opened 2018-03-14
TSK-386 rest interfaces for domain query in admin

LarsLeo:
rest services for domain query


MiguelMartinRojas:
I will close it because Lars will push a new PR with UI and Services.

PR #294, opened 2018-03-19
Tsk217: pagination support for WorkbasketController

holgerhagen:


rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketController.java
```diff
@@ -377,4 +386,19 @@ private void addAttributeFilter(WorkbasketQuery query,
             }
         }
     }
+
+    private PageMetadata initPageMetadata(String pagesizeParam, String pageParam, long totalElements)
+        throws InvalidArgumentException {
+        long pagesize;
```
MiguelMartinRojas: TYPO: use camecase notation pageSize

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketController.java
```diff
@@ -93,18 +93,27 @@
...
         addSortingToQuery(query, sortBy, order);
         addAttributeFilter(query, name, nameLike, key, keyLike, descLike, owner, ownerLike, type);
         addAuthorizationFilter(query, requiredPermission);
 
-        // long totalElements = query.count();
-
-        PageMetadata pageMetadata = new PageMetadata(Integer.MAX_VALUE, 0, 11, 4);
-        List<WorkbasketSummary> workbasketSummaries = query.listPage((int) pageMetadata.getNumber(),
-            (int) pageMetadata.getSize());
+        PageMetadata pageMetadata = null;
```
MiguelMartinRojas: I would add all this pagination logic into a function. AddPagingToQuery or something like this.

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/resource/mapper/WorkbasketSummaryResourcesAssembler.java
```diff
@@ -38,7 +38,22 @@ public WorkbasketSummaryResourcesAssembler() {
 
         UriComponentsBuilder original = getBuilderForOriginalUri();
         pagedResources.add(new Link(original.toUriString()).withSelfRel());
-        pagedResources.add(linkTo(WorkbasketController.class).withRel("allWorkbaskets"));
+        if (pageMetadata != null) {
```
MiguelMartinRojas: I didn't expect to get previous, next, first and last pages, It's nice having that :)

holgerhagen: I thought this would be nice for our HATEOAS concept. But you are probably right and I overengineered this part a little bit... :-)

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketController.java
holgerhagen: I took this from an example. I believe you're right.

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketController.java
holgerhagen: I considered this. But this part actually also contains the non-paging part. Therefore this does not work. Since I expect a larger refactoring as soon as we do have another paging resource I would think about a better design at that point in time.

PR #304, opened 2018-03-23
TSK-295

LarsLeo:
created import/export components
also includes TSK-386
except for reformatting and grammar correction, there were no changes in app-component files
also just small changes in workbasket-list.component.ts and .angular-cli.json, workbasket.service


MiguelMartinRojas:
Lars YouTube need to run nun build:prod before pushing your chances because It has linting check.

PR #311, opened 2018-03-28
TSK-416

LarsLeo:
now using generic domains/ did refactors of import/export

### DIFF
--- a/web/src/app/administration/classification/master/list/classification-list.component.ts
+++ b/web/src/app/administration/classification/master/list/classification-list.component.ts
@@ -1,4 +1,5 @@
 import {Component, OnInit} from '@angular/core';
+import {SelectionToImport} from '../../../enums/SelectionToImport';
 
 @Component({
   selector: 'taskana-classification-list',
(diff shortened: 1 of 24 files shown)


web/src/app/administration/classification/master/list/classification-list.component.ts
```diff
@@ -1,4 +1,5 @@
 import {Component, OnInit} from '@angular/core';
+import {SelectionToImport} from '../../../enums/SelectionToImport';
```
MiguelMartinRojas: I am using app/enums/selectionToImport  instead of  parent navigation.

web/src/app/administration/enums/SelectionToImport.ts
```diff
@@ -0,0 +1,4 @@
+export enum SelectionToImport {
```
MiguelMartinRojas: I prefer to move this clases to models folder.

web/src/app/services/domains/domain.service.ts
```diff
@@ -0,0 +1,26 @@
...
+      'Authorization': 'Basic VEVBTUxFQURfMTpURUFNTEVBRF8x'
+    })
+  };
+
+  constructor(private httpClient: HttpClient) {
+  }
+
+  // TODO how to enable hateoas support?
+  // GET
+  getDomains(): Observable<string[]> {
```
MiguelMartinRojas: It needs to be implemented on server side.

PR #315, opened 2018-04-03
TSK-399: Configurable classification cagetories.

holgerhagen:


lib/taskana-core/src/main/java/pro/taskana/configuration/TaskanaEngineConfiguration.java
```diff
@@ -73,6 +74,9 @@
     // List of configured classification types
     protected List<String> classificationTypes = new ArrayList<String>();
 
+    // List of configured classification categories
+    protected List<String> classificationCategories = new ArrayList<String>();
+
```
BerndBreier: classificationTypes and classificationCategories should better be Set s to prevent duplicate entries...

PR #330, opened 2018-04-11
Tsk 204 205 206 208 classification actions

MiguelMartinRojas:
- Add new classification.
- Reset classification fields.
- Delete classification (REST issues)
- Delete classification end point (Issues).
- Fix IE 10 bugs (Spinner, forms hidden...).
- Classification tree deep linking.
- Expand classification tree on deep linking.


MiguelMartinRojas:
Ok, I will add this option

PR #338, opened 2018-04-16
Task/431 now releasing web app as maven artifact

mustaphazorgati:


web/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+      <name>Holger Hagen</name>
+      <email>holger.hagen@novatec-gmbh.de</email>
+      <organization>NovaTec Consulting GmbH</organization>
+      <organizationUrl>https://www.novatec-gmbh.de</organizationUrl>
+    </developer>
+  </developers>
+
+  <parent>
+    <groupId>org.springframework.boot</groupId>
+    <artifactId>spring-boot-starter-parent</artifactId>
```
TitusLabs: this is such a minimal jar, we do not need a parent - at least spring boot - here..

web/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+                  <directory>dist</directory>
+                </resource>
+              </resources>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-source-plugin</artifactId>
```
TitusLabs: we do not need a source jar since the jar itself is already a source jar ...or you need to package the web sources this is build from - but I think this not needed

web/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+            <id>attach-sources</id>
+            <goals>
+              <goal>jar-no-fork</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-javadoc-plugin</artifactId>
```
TitusLabs: we have no java docs in the web app

PR #352, opened 2018-04-19
TSK-453 Add external configuration file for environment setup

MiguelMartinRojas:
- Added external configuration file for url setup.
- If urls are empty string or file is not created, default url configuration is used
- Configuration file is located at: web/src/environments/data-sources/environment-information.json
- To change the file after AOT compilation file is located at \web\dist\environments\data-sources\*

### DIFF
--- a/web/.angular-cli.json
+++ b/web/.angular-cli.json
@@ -9,7 +9,8 @@
 			"outDir": "dist",
 			"assets": [
 				"assets",
-				"taskana.ico"
+				"taskana.ico",
+				"environments/data-sources"
 			],
 			"index": "index.html",
 			"main": "main.ts",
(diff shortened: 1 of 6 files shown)


TitusLabs:
OK, I think this should work.. but, there is a chance that my productive environment could point to the demo-environment due to the default values.
After all, we should also change the deployment to not deliver the demo-urls within the jar file..
What do you think?

MiguelMartinRojas:
I you want to make sure that the jar file doesn't contains any demo-enviroment urls we could add a third environment  property when building  angular application, currently we have dev: localhost and prod: blue-mix demo, but we could rename prod to demo and add a third one called prod with empty urls, this way prod environment will only can be setup by external config file.

TitusLabs:
Yes, I think this could be one solution. I will open a ticket for this and we can diskuss this afterwards. It is not critical for this release..

PR #389, opened 2018-05-14
TSK-336: Workplace Prototype

LarsLeo:


web/src/app/shared/master-and-detail/master-and-detail.component.ts
```diff
@@ -11,7 +11,8 @@ import { MasterAndDetailService } from 'app/services/masterAndDetail/master-and-
 export class MasterAndDetailComponent implements OnInit {
     private classifications = 'classifications';
     private workbaskets = 'workbaskets';
```
MiguelMartinRojas: Use const for constants

web/src/app/workplace/services/task.service.ts
```diff
@@ -1,18 +1,26 @@
-import { Task } from '../models/task';
-import { Observable } from 'rxjs/Observable';
-import { HttpClient, HttpHeaders } from '@angular/common/http';
-import { Injectable } from '@angular/core';
-import { environment } from 'app/../environments/environment';
+import {Task} from '../models/task';
```
MiguelMartinRojas: Please use .app/models/tasks for importing path

web/src/app/workplace/services/task.service.ts
```diff
@@ -1,18 +1,26 @@
...
+import {Observable} from 'rxjs/Observable';
+import {HttpClient, HttpHeaders} from '@angular/common/http';
+import {Injectable} from '@angular/core';
+import {environment} from 'app/../environments/environment';
 
 @Injectable()
 export class TaskService {
   url = environment.taskanaRestUrl + '/v1/tasks';
 
+  httpOptions = {
```
MiguelMartinRojas: Remove authorization headers from services, I have created an interceptor for share module, and workplace module which overrides the default one.
[This applies to all authorization headers hardcoded]

web/src/app/workplace/services/task.service.ts
```diff
@@ -1,18 +1,26 @@
...
-import { environment } from 'app/../environments/environment';
+import {Task} from '../models/task';
+import {Observable} from 'rxjs/Observable';
+import {HttpClient, HttpHeaders} from '@angular/common/http';
+import {Injectable} from '@angular/core';
+import {environment} from 'app/../environments/environment';
 
 @Injectable()
 export class TaskService {
   url = environment.taskanaRestUrl + '/v1/tasks';
```
MiguelMartinRojas: Please use template string for all strings with params, It's very usefull feature from modern programming languages, https://basarat.gitbooks.io/typescript/docs/template-strings.html
[This comment apply to all strings].

web/src/app/workplace/task/task.component.ts
```diff
@@ -13,29 +13,28 @@ import { WorkbasketService } from '../services/workbasket.service';
   styleUrls: ['./task.component.scss']
 })
 export class TaskComponent implements OnInit {
-
   task: Task = null;
-  link: SafeResourceUrl = this.sanitizer.bypassSecurityTrustResourceUrl('https://duckduckgo.com/?q=');
+  link: SafeResourceUrl = this.sanitizer.bypassSecurityTrustResourceUrl('https://bing.com');
```
MiguelMartinRojas: move https://bing.com to a const

web/src/app/workplace/tasklist/tasklist.component.scss
```diff
@@ -1,3 +1,52 @@
 .clickable {
     cursor: pointer;
 }
+
+.task-list-full-height{
```
MiguelMartinRojas: We are using .workbasket-list-full-height, task-list-full-height and workbasket-list-full-height with the same rule in three different parts of the code, maybe we can extract this into master-and-detail.component, probably should extract all common  sass into parent sass

web/src/app/workplace/workplace-routing.module.ts
```diff
@@ -1,36 +1,44 @@
...
-        path: 'tasks/taskdetail/:id',
-        component: TaskdetailsComponent
-      }
-    ]
-  }
-];
+    {
+      path: 'tasks',
+      component: MasterAndDetailComponent,
+      canActivate: [DomainGuard],
```
MiguelMartinRojas: Do you need DomainGuard for tasks route?

web/src/app/workplace/task/task.component.ts
```diff
@@ -13,29 +13,28 @@ import { WorkbasketService } from '../services/workbasket.service';
...
   ngOnInit() {
+    this.requestInProgress = true;
     const id = this.route.snapshot.params['id'];
     this.taskService.getTask(id).subscribe(
       t => {
         this.task = t;
-        this.link = this.sanitizer.bypassSecurityTrustResourceUrl('https://duckduckgo.com/?q=' + this.task.name);
+        this.link = this.sanitizer.bypassSecurityTrustResourceUrl('https://bing.com/?q=' + this.task.name);
         this.workbasketService.getAllWorkBaskets().subscribe(w => {
           this.workbaskets = w['_embedded']['workbaskets'];
```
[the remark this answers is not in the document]
LarsLeo: when i do it this way, i get a "property does not exist: _embedded"

web/src/app/workplace/taskdetails/taskdetails.component.ts
```diff
@@ -11,24 +11,37 @@ import {Location} from '@angular/common';
...
   constructor(private route: ActivatedRoute,
               private taskService: TaskService,
-              private location: Location) {
+              private router: Router) {
   }
 
   ngOnInit() {
-    this.getTask();
+    this.routeSubscription = this.route.params.subscribe(params => {
+      this.requestInProgress = true;
```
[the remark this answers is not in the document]
LarsLeo: why exactly is this better? Should it not be the same?

web/src/app/workplace/task/task.component.ts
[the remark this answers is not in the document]
MiguelMartinRojas: Yes, you need to check  first if workbasket has embedded property created.

web/src/app/shared/master-and-detail/master-and-detail.component.ts
LarsLeo: i don't think you can use constants for class members

web/src/app/workplace/taskdetails/taskdetails.component.ts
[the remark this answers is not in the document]
MiguelMartinRojas: Because requestInProgress is for async calls, if you follow the flow of this two functions you will find the next execution order, since taskService.getTask(id)  is returning an observable, so You are setting requestInProgress to false before tasks are being retrieved.

this.routeSubscription = this.route.params.subscribe(params => {
[1]
    this.requestInProgress = true;
     const id = params['id'];
     this.getTask(id);
[3]
      this.requestInProgress = false;
   });

getTask(id: string): void {
[2]
     this.taskService.getTask(id).subscribe(task => {
[4]
    this.task = task;   });
 }

------------------ correct way

getTask(id: string): void {
this.requestInProgress = true;
     this.taskService.getTask(id).subscribe(task => {
     this.requestInProgress = false;
     this.task = task;   });
 }

web/src/app/workplace/workplace-routing.module.ts
MiguelMartinRojas: Guards in angular are use to allow/forbidden access to routes depending on conditions, you don't need any domain check, I believe...

web/src/app/shared/master-and-detail/master-and-detail.component.ts
MiguelMartinRojas: forgive this comment... but constant can be declared outside of the class. But don't do It, I will do and apply some refactor for that.

PR #397, opened 2018-05-22
TSK-505 shared WorkbasketService

LarsLeo:


web/src/app/services/workbasket/workbasket.service.ts
```diff
@@ -92,6 +96,12 @@ export class WorkbasketService {
 	getWorkBasket(id: string): Observable<Workbasket> {
 		return this.httpClient.get<Workbasket>(`${environment.taskanaRestUrl}/v1/workbaskets/${id}`);
 	}
+
+	// GET
+  getAllWorkBaskets(): Observable<WorkbasketResource> {
```
MiguelMartinRojas: Use getWorkBasketsSummary method with requiredPermission property and make the others null in the call.

PR #400, opened 2018-05-23
TSK-483-508 make lookup configurable and refactored environment file

MiguelMartinRojas:
- lookup field is configurable via configuration file. ¨customized-fields.json¨ 
    if visible = true => lookup field will be used else => input field will be used
       "lookupField": {
                    "visible": true
       },

- Moved customized-fields configuration into a different config file, as Titus requested on TSK-483


TitusLabs:
is there also a default language? so, what happens if the configured language in the browser is not present in the configurations?

web/src/app/administration/workbasket/details/information/workbasket-information.component.html
```diff
@@ -39,7 +39,12 @@ <h4 class="panel-header">{{workbasket.name}}&nbsp;
                 </div>
                 <div class="form-group required">
                     <label for="wb-owner" class="control-label">Owner</label>
-                    <taskana-type-ahead required #owner="ngModel" name="owner" [(ngModel)]="workbasket.owner" placeHolderMessage="Owner is required"></taskana-type-ahead>
+                    <taskana-type-ahead *ngIf="lookupField.visible else ownerInput" required #owner="ngModel" name="owner" [(ngModel)]="workbasket.owner"
+                        placeHolderMessage="Owner is required"></taskana-type-ahead>
```
TitusLabs: the access right owner is ment here !?

web/src/environments/data-sources/titles.json
```diff
@@ -1,18 +1,16 @@
...
-        }
+    "EN": {
+        "0": "Administration",
+        "1": "Custom 1",
+        "2": "Custom 2",
+        "3": "Custom 3",
+        "4": "Custom 4"
+    },
+    "DE": {
+        "0": "Verwaltung",
```
TitusLabs: this could be called Administration too

web/src/environments/data-sources/titles.json
```diff
@@ -1,18 +1,16 @@
...
+        "1": "Gewohnheit 1",
```
TitusLabs: hmm, tough one.. if not custom too... i would call it for example "Freifeld" !?

TitusLabs: or i would prefer this "Zusatzfeld"

holgerhagen: I like Zusatzfeld

web/src/app/administration/workbasket/details/information/workbasket-information.component.html
MiguelMartinRojas: What do you mean?

TitusLabs: never mind, i was not sure about the context here.. i think it is ok

PR #401, opened 2018-05-23
TSK-502 Make Taskana run on DB2 10.5

BerndBreier:


lib/taskana-core/src/main/java/pro/taskana/mappings/QueryMapper.java
```diff
@@ -32,7 +32,14 @@
         + "<where>"
         + "<if test='accessIdIn != null'> "
         + "AND t.WORKBASKET_ID IN ( "
+        + "<choose>"
+        + "<when test=\"_databaseId == 'db2'\">"
```
mustaphazorgati: The explicit cast from bool -> int was introduced for the postgresql and is only necessary for the postgresql. Instead of not casting only for the db2 instance i'd rather only cast when a postgres db is found.
This does not change the outcome, but I think that it will make it a little clearer why we decided to handle it this way. [ Furthermore the extra processing (even if it's very little) will only be done from the postgresql and not from the other databases. ]

lib/taskana-core/src/test/java/pro/taskana/database/TestDataGenerator.java
```diff
@@ -43,6 +44,9 @@ public void generateTestData(DataSource dataSource) throws SQLException {
         ScriptRunner runner = null;
         try {
             Connection connection = dataSource.getConnection();
+            if (connection.getMetaData().getDatabaseProductName().contains("DB2")) {
```
mustaphazorgati: Isn't it smarter to create a static method TaskanaEngineImpl.isDb2(String dbProductName) so that the identification of a db2 database is only located in one space?

lib/taskana-core/src/test/java/pro/taskana/database/TestDataGenerator.java
```diff
@@ -51,12 +55,21 @@ public void generateTestData(DataSource dataSource) throws SQLException {
             runner.setStopOnError(true);
             runner.setLogWriter(this.logWriter);
             runner.setErrorLogWriter(this.errorLogWriter);
-            runner.runScript(
-                new InputStreamReader(this.getClass().getResourceAsStream("/sql/classification.sql")));
-            runner.runScript(new InputStreamReader(this.getClass().getResourceAsStream("/sql/workbasket.sql")));
-            runner.runScript(new InputStreamReader(this.getClass().getResourceAsStream("/sql/task.sql")));
-            runner.runScript(
-                new InputStreamReader(this.getClass().getResourceAsStream("/sql/workbasket-access-list.sql")));
+            if (isDb2) {
```
mustaphazorgati: I'd prefer to replace false/true to 0/1 in memory instead of creating extra sql files for that. This way we don't have to worry about inconsistency about those files. generateMonitoringSqlData() is a good example. 
A step further would be to save those Strings in static variables so that the replacement is only necessary once. This assumes that the datasource does not change in one run (which it does not)

lib/taskana-core/src/main/java/pro/taskana/mappings/QueryMapper.java
holgerhagen: we should make the more unusual case the special case. in this case db2, because boolean is supported by most databases and will be in DB2 11, won't it? This way we can remove the special case easier later on...

lib/taskana-core/src/test/java/pro/taskana/database/TestDataGenerator.java
holgerhagen: I agree. Don't you have it available already?

lib/taskana-core/src/main/java/pro/taskana/mappings/QueryMapper.java
mustaphazorgati: Yes, all our other dbs support the (::) cast. I just did a quick research and didn't figure out which dbs support that cast. As far as I remember i'd say that most of them do (don't take my word on this one)   

So yes, we can leave it as it is in order to remove it easier.

lib/taskana-core/src/test/java/pro/taskana/database/TestDataGenerator.java
mustaphazorgati: Yes. I have already implemented it. I'll add a commit to this PR in a little.

BerndBreier: Ok, so I don't do anything more with this request. I guess, you will also add the static method TaskanaEngineImpl.isDb2(). If you are explicitely asking whether it's DB2 you don't need a parameter on this method ...

BerndBreier: By the time when you call the isDb2() method, TaskanaEngineImpl has already been initialized and especially, createSqlSessionManager() has already been executed. This method determines the database system and could save either the db system name or a 'isDb2' flag statically.
But it is prudent to not assume this order of execution and pass a connection to the isDB2() method. In this way you are on the safe side ...

PR #406, opened 2018-05-25
TSK-493 Check classification Id is null on creation

MiguelMartinRojas:
- Check classification id is null on classification creation.


rest/taskana-rest-spring-example/src/test/java/pro/taskana/rest/ClassificationControllerIntTest.java
```diff
@@ -126,7 +130,7 @@ public void testGetSecondPageSortedByKey() {
 
     @Test
     public void testCreateClassification() throws IOException {
-        String newClassification = "{\"classificationId\":\"\",\"category\":\"MANUAL\",\"domain\":\"DOMAIN_A\",\"key\":\"NEW_CLASS\",\"name\":\"new classification\",\"type\":\"TASK\"}";
+        String newClassification = "{\"classificationId\":null,\"category\":\"MANUAL\",\"domain\":\"DOMAIN_A\",\"key\":\"NEW_CLASS\",\"name\":\"new classification\",\"type\":\"TASK\"}";
```
holgerhagen: Empty classification id is also valid, isn't it? Why did you change the code here?

holgerhagen: We should also have a test returning a HTTP 400 if ID is set.

lib/taskana-core/src/test/java/pro/taskana/impl/ClassificationServiceImplTest.java
```diff
@@ -290,19 +291,38 @@ public void testGetClassificationWithInvalidNullKey() throws ClassificationNotFo
         }
     }
 
+    @Test(expected = InvalidArgumentException.class)
+    public void testGetClassificationWithInvalidNullId()
```
holgerhagen: this should probably be named like "testThrowExceptionIdClassificationIsCreatedWithAnExplicitId" or similar. What do you think?

MiguelMartinRojas: Yes, you are right your name is more descriptive.

rest/taskana-rest-spring-example/src/test/java/pro/taskana/rest/ClassificationControllerIntTest.java
MiguelMartinRojas: ok

PR #408, opened 2018-05-28
TSK-535: Bugfixes for Workplace

LarsLeo:
lots of bugfixes on the workplace, also renamed a PathVariable in the TaskController transfer method

### DIFF
--- a/web/src/app/workplace/taskdetails/taskdetails.component.ts
+++ b/web/src/app/workplace/taskdetails/taskdetails.component.ts
@@ -1,15 +1,16 @@
-import {Component, OnInit} from '@angular/core';
-import {Task} from '../models/task';
+import {Component, OnDestroy, OnInit} from '@angular/core';
+import {Task} from 'app/workplace/models/task';
 import {ActivatedRoute, Router} from '@angular/router';
-import {TaskService} from '../services/task.service';
+import {TaskService} from 'app/workplace/services/task.service';
 import {Subscription} from 'rxjs/Subscription';
+import {Location} from '@angular/common';
 
 @Component({
   selector: 'taskana-task-details',
   templateUrl: './taskdetails.component.html',
   styleUrls: ['./taskdetails.component.scss']
 })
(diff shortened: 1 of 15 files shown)


web/src/app/workplace/taskdetails/taskdetails.component.ts
```diff
@@ -46,4 +46,12 @@ export class TaskdetailsComponent implements OnInit {
   openTask(taskId: string) {
     this.router.navigate([{outlets: {detail: `task/${taskId}`}}], {relativeTo: this.route.parent});
   }
+
+  workOnTaskDisabled(): boolean {
+    return this.task === null ? false : this.task.state === 'COMPLETED';
```
MiguelMartinRojas: Do not compare with null in typescript or javascript, there is no need for that and some bugs can come from that
return this.task? this.task.state === 'COMPLETED' : false ;

web/src/app/workplace/workbasket-selector/workbasket-selector.component.ts
```diff
@@ -61,4 +56,9 @@ export class SelectorComponent implements OnInit {
         tasks._embedded.tasks.forEach(e => this.tasks.push(e));
       });
   }
+
+  isDisabled(): boolean {
```
MiguelMartinRojas: this is not efficient, You are iterating over an array everytime something in the UI is refreshed.
You need to explain me what you want to do here.

LarsLeo: well i want the button to only be active, when the input field contains one of the actual workbasket names, since there is no reason to search for a non existing workbasket

MiguelMartinRojas: You are using, component "ng2-auto-complete": "0.12.0" for this... We have already imported ng-bootstrap components library, and you can use typeahead component which has already implemented onSelect(). You can use this onSelect() function to check if the workbasket is correct or not, simply using if(selected) then is correct else wrong ...

PR #435, opened 2018-06-14
TSK-516 verify schema version at startup

BerndBreier:


lib/taskana-core/src/main/java/pro/taskana/configuration/DbSchemaCreator.java
```diff
@@ -76,6 +78,39 @@ private boolean isSchemaPreexisting(ScriptRunner runner) {
...
     }
 
+    public boolean isValidSchemaVersion(String expectedVersion) {
+        SqlRunner runner = null;
+        try {
+            Connection connection = dataSource.getConnection();
+            runner = new SqlRunner(connection);
+            LOGGER.debug(connection.getMetaData().toString());
+
+            String query = "select VERSION from TASKANA.TASKANA_SCHEMA_VERSION where ID = '1' AND VERSION = '"
```
holgerhagen: I believe this is wrong. I would expect a list of versions later on (each fix and upgrade leaving its signature) - therefore we should query for the largest number and compare that one.

lib/taskana-spring/src/main/java/pro/taskana/configuration/SpringTaskanaEngineConfiguration.java
```diff
@@ -1,53 +1,63 @@
...
+    /**
+     * This method creates the Spring-based TaskanaEngine without an sqlSessionFactory
+     *
+     * @return the TaskanaEngine
+     */
+    @Override
+    public TaskanaEngine buildTaskanaEngine() {
+        this.useManagedTransactions = true;
+
+        dbSchemaCreator = new DbSchemaCreator(this.dataSource);
```
holgerhagen: this is duplicate code. why do we run the schema creator in the constructor for TaskanaEngineConfiguration and in buildTaskanaEngine in SpringTaskanaEngineConfiguration? I would prefer to have the schema creation in the superclass and remove it from the Spring derivation.

PR #438, opened 2018-06-15
TSK-559: Improved test coverage for the workbasket mapper

JulianSchallenmueller:


lib/taskana-core/src/test/java/acceptance/workbasket/CreateWorkbasketAccTest.java
```diff
@@ -57,6 +58,9 @@ public void testCreateWorkbasket()
         Workbasket createdWorkbasket = workbasketService.getWorkbasket("NT1234", "DOMAIN_A");
         assertNotNull(createdWorkbasket);
         assertNotNull(createdWorkbasket.getId());
+
+        Assert.assertEquals(workbasket, createdWorkbasket);
```
holgerhagen: "Assert." is not needed. See other statements. 

What are the blank lines for?

JulianSchallenmueller: This part is now adjusted with the blank lines removed, i'll wait for new information on the method mentioned above before updating this PR.

PR #442, opened 2018-06-18
TSK-528 now testing multiple databases simultaneously

mustaphazorgati:


mustaphazorgati:
I might have to encrypt the ci/codesigning.asc.enc file again because travis was not able to decrypt it properly.

PR #444, opened 2018-06-18
TSK-567 Create a scheduler job for cleanup completed tasks

MiguelMartinRojas:


lib/taskana-core/src/main/java/pro/taskana/TaskanaEngine.java
```diff
@@ -66,6 +67,26 @@
      */
     void setConnection(java.sql.Connection connection) throws SQLException;
 
+
+    /**
+     * Open the connection to the database. to be called at the begin of each Api call that accesses the database
+     */
+    void openConnection();
+
```
BerndBreier: openConnection, getSqlSession and returnConnection are calls that are not meant as public API. These calls are used by Taskana Implementation classes. Opening them for use by the customer may cause problems because usage has to adhere to a defined protocol.

lib/taskana-core/src/main/java/pro/taskana/TaskanaEngine.java
```diff
@@ -66,6 +67,26 @@
...
+    /**
+     * retrieve the SqlSession used by taskana.
+     *
+     * @return the myBatis SqlSession object used by taskana
+     */
+    SqlSession getSqlSession();
```
BerndBreier: See above

lib/taskana-core/src/main/java/pro/taskana/TaskanaEngine.java
```diff
@@ -66,6 +67,26 @@
...
+
+    /**
+     * Returns the database connection into the pool. In the case of nested calls, simply pops the latest session from
+     * the session stack. Closes the connection if the session stack is empty. In mode AUTOCOMMIT commits before the
+     * connection is closed. To be called at the end of each Api call that accesses the database
+     */
+    void returnConnection();
```
BerndBreier: See above. This call should not be publicly available.

lib/taskana-core/src/main/java/pro/taskana/impl/TaskanaEngineImpl.java
```diff
@@ -222,7 +222,7 @@ public void closeConnection() {
     /**
      * Open the connection to the database. to be called at the begin of each Api call that accesses the database
      */
-    void openConnection() {
+    public void openConnection() {
```
BerndBreier: openConnection is an internal call that should not be public.

lib/taskana-core/src/main/java/pro/taskana/impl/TaskanaEngineImpl.java
```diff
@@ -245,7 +245,7 @@ void initSqlSession() {
      * the session stack. Closes the connection if the session stack is empty. In mode AUTOCOMMIT commits before the
      * connection is closed. To be called at the end of each Api call that accesses the database
      */
-    void returnConnection() {
+    public void returnConnection() {
```
BerndBreier: returnConnection is an internal call that should not be public.

lib/taskana-core/src/main/java/pro/taskana/impl/TaskanaEngineImpl.java
```diff
@@ -267,7 +267,7 @@ void returnConnection() {
      *
      * @return the myBatis SqlSession object used by taskana
      */
-    SqlSession getSqlSession() {
+    public SqlSession getSqlSession() {
```
BerndBreier: getSqlSession is an internal call that should not be public.

lib/taskana-core/src/main/java/pro/taskana/impl/jobs/JobTaskRunner.java
```diff
@@ -0,0 +1,60 @@
+package pro.taskana.impl.jobs;
+
```
BerndBreier: create JobTasksRunner in package pro.taskana.impl, then TaskanaEngineImpl.openConnection, returnConnection etc don't need to be public.

lib/taskana-core/src/main/java/pro/taskana/impl/jobs/JobTaskRunner.java
```diff
@@ -0,0 +1,60 @@
...
+            List<TaskSummary> tasksCompleted = taskService.createTaskQuery()
+                .completedWithin(new TimeInterval(null, untilDate))
+                .list();
+            tasksCompleted.stream().forEach(task -> {
+                tasksIds.add(task.getTaskId());
+                LOGGER.info("task id deleted: {}", task.getTaskId());
+            });
+            bulkLog = taskService.deleteTasks(tasksIds);
+        } catch (Exception e) {
+            taskanaEngine.returnConnection();
```
BerndBreier: where is the associated openConnection() call? These calls must always occur in pairs.
And why is returnConnection called only in the catch() branch?

lib/taskana-core/src/test/java/acceptance/classification/UpdateClassificationAccTest.java
```diff
@@ -28,7 +28,7 @@
 import pro.taskana.exceptions.NotAuthorizedException;
 import pro.taskana.exceptions.TaskNotFoundException;
 import pro.taskana.impl.DaysToWorkingDaysConverter;
-import pro.taskana.impl.JobRunner;
+import pro.taskana.impl.jobs.JobRunner;
```
BerndBreier: leave JobRunner in the pro.taskana.impl package. This removes the necessity to make openConnection(), returnConnection() and getSqlSession() public.

rest/taskana-rest-spring-example/src/main/java/pro/taskana/rest/JobScheduler.java
```diff
@@ -22,9 +30,13 @@
...
+    private final long untilDays = 14;
+
     private static final Logger LOGGER = LoggerFactory.getLogger(JobScheduler.class);
     @Autowired
     private TaskanaEngine taskanaEngine;
+    @Autowired
+    private TaskService taskService;
 
     @Scheduled(fixedRate = 60000)
     @Transactional(rollbackFor = Exception.class)
```
BerndBreier: This @Transactional attribute must be removed. Otherwise, transaction slicing as introduced with TSK-511 would not work.

BerndBreier:
With the changes included, JobScheduler.triggerTaskCompletedCleanUpJob has no @Transactional annotation, and it doesn't use the TransactionProvider.executeInTransaction() idiom to control transactions. Ergo: there is no transaction active when jobs are deleted and as a result, jobs will not be deleted....
I would strongly recommend to use the TransactionProvider.executeInTransaction() pattern and to split the deletion in batches of defined size, because after 2 weeks there may be very many tasks around and an attempt to delete them all in one transaction could take longer than the transaction timeout causing the complete transaction to roll back ....

PR #450, opened 2018-06-20
TSK-563: Improved task mapper test coverage

JulianSchallenmueller:


JulianSchallenmueller:
Okey, i'll put it back in then.

PR #457, opened 2018-06-21
TSK-572: fixed NPE in WorkbasketServiceImpl

holgerhagen:


lib/taskana-core/src/main/java/pro/taskana/impl/WorkbasketServiceImpl.java
```diff
@@ -411,8 +412,11 @@ public WorkbasketAccessItem updateWorkbasketAccessItem(WorkbasketAccessItem work
         List<WorkbasketPermission> permissions = new ArrayList<>();
         WorkbasketAccessItem wbAcc = workbasketAccessMapper.findByWorkbasketAndAccessId(workbasketId,
             CurrentUserContext.getAccessIds());
-        this.addWorkbasketAccessItemValuesToPermissionSet(wbAcc, permissions);
+        if (wbAcc != null) {
```
MiguelMartinRojas: I prefer using null check in avance and return the default value:
if (wbAcc == null){
  return permissions
}
return  this.addWorkbasketAccessItemValuesToPermissionSet(wbAcc);

Also I would delegate the creation of the List<WorkbasketPermission>permissions = new ArrayList<>(); to  addWorkbasketAccessItemValuesToPermissionSet and use return function value instead of using an output permissions parameter.

MiguelMartinRojas:
Also remove default constructor. It's not been used and It should be used.

PR #460, opened 2018-06-21
TSK-587 Improve Job Support

BerndBreier:


rest/taskana-rest-spring-example/src/main/resources/application.properties
```diff
@@ -26,5 +26,7 @@ taskana.ldap.groupSearchFilterValue=groupOfUniqueNames
 taskana.ldap.groupNameAttribute=cn
 taskana.ldap.minSearchForLength=3
 taskana.ldap.maxNumberOfReturnedAccessIds=50
+####### JobScheduler delay between runs in milliseconds
```
holgerhagen: This is not true anymore, is it?
BTW: what is the minimum time between two starts when we are using cron? Is it possible to start every second if we would like to?

rest/taskana-rest-spring-example/src/test/java/pro/taskana/rest/ClassificationControllerIntTest.java
```diff
@@ -273,9 +279,31 @@ public void testUpdateClassificationPrioServiceLevel()
...
+            }
+
+            @Override
+            public Date lastCompletionTime() {
+                return null;
+            }
+        };
+        long delay = trigger.nextExecutionTime(context).getTime() - (new Date()).getTime() + 2000;
+
+        LOGGER.info("About to sleep for " + delay / 1000
```
holgerhagen: This is not working anymore, is it?

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
```diff
@@ -1394,14 +1394,14 @@ private void initAttachment(AttachmentImpl attachment, Task newTask) {
         }
     }
 
-    BulkOperationResults<String, Exception> classificationChanged(String taskId, String classificationId)
+    BulkOperationResults<String, Exception> refreshPriorityAndDueDate(String taskId, String modifiedClassificationId)
```
holgerhagen: Why do we need the classificationId in the signature? The method would make more sense to me if we would remove it.

rest/taskana-rest-spring-example/src/main/resources/application.properties
BerndBreier: Ok, I overlooked the comment. A cron expression for Spring has 6 parts:
<second> <minute> <hour> <day-of-month> <month> <day-of-week>
we set <second> to 0 and all other to '*'. This means, the expression triggers each full minute.
If you set also second to '*', the scheduler would run every second.

rest/taskana-rest-spring-example/src/test/java/pro/taskana/rest/ClassificationControllerIntTest.java
BerndBreier: Yes, it is working. It parses the cron expression and uses the result to calculate the delay up to the next run of JobScheduler ...

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
BerndBreier: Good point - we don't need this parameter really. I'll remove it.

BerndBreier: Thanks god, I removed the classificationId parameter. In this way I found an error in method refreshPriorityAndDueDate(). It contained the 2 lines:

            Classification classification = classificationService.getClassification(modifiedClassificationId);
            task.setClassificationSummary(classification.asSummary());

This is an error because the task to modify might have a completely different classification and only be affected because one of its attachments had the updated classification.

rest/taskana-rest-spring-example/src/main/resources/application.properties
BerndBreier: Hmm, in the previous append some content has been lost...
The correct explanation should be:
A cron expression for Spring has 6 parts:
second minute hour day-of-month month day-of-week
when we specify  0 * * * * *, we set second to 0 and all other to asterisk. This means, the expression triggers each full minute.
If you set also second to asterisk, the scheduler would run each second.

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
BerndBreier: I think, the test coverage is fair because we modify a classification that is widely used. We set prio to 1000 and SL to P15D. The result:
37 tasks are directly affected because the use this classification directly. 36 from these 37 don't have attachments, therefore they get the new prio / SL from the classification. 1 from these 37 has also an attachment with SL P8D, therefore it gets a due date of planned = 8 working days.
4 tasks are affected only indirectly because only an attachment uses the modified classification.
Since these tasks use a classification with SL P1D, they end up with a due date of planned + 1 working day.
I double checked the expected results and found that in the previous version, the wrong results were expected. Now everything should be correct.
I added the check for correct prio / due date for the indirectly affected tasks to the testcase

rest/taskana-rest-spring-example/src/test/java/pro/taskana/rest/ClassificationControllerIntTest.java
MiguelMartinRojas: I have noticed that rest tests are sleeping for sometime(1,5min) to process the job execution.
I suggest to make cron configurable this way test could configure job schedule to less time(5 sec?) and tests will run faster.

BerndBreier: This is exactly one test: ClassificationControllerIntTest.testUpdateClassificationPrioServiceLevel().
With this pull request, this test parses the cron expression that controls JobScheduler and calculates the delay between now and the next trigger time. Then it sleeps for 2 seconds more that that delay...

MiguelMartinRojas: ok, perfect.

PR #470, opened 2018-06-27
TSK-608: removed exception hiding in JobRunner.

holgerhagen:


lib/taskana-core/src/main/java/pro/taskana/impl/JobRunner.java
```diff
@@ -119,6 +119,7 @@ private void processAJob(BulkOperationResults<String, Exception> bulkLog, Job jo
             }
         } catch (Exception e) {
             // transaction was rolled back -> split job into 2 half sized jobs
+            LOGGER.warn("Processing ob job " + job.getJobId() + " failed. Trying to split it up into two pieces...", e);
```
MiguelMartinRojas: There is a typo, processing OF job

PR #490, opened 2018-07-04
TSK-626: Add Channel and ObjectReference to AttachmentSummary

BVier:


lib/taskana-core/src/main/java/pro/taskana/impl/AttachmentSummaryImpl.java
```diff
@@ -176,13 +207,27 @@ public boolean equals(Object obj) {
...
+        } else if (!channel.equals(other.channel)) {
+            return false;
+        }
         return true;
     }
 
     @Override
     public String toString() {
         StringBuilder builder = new StringBuilder();
-        builder.append("AttachmentSummaryImpl [id=");
```
TitusLabs: why? this is the summary..

lib/taskana-core/src/test/java/acceptance/task/QueryTaskWithAttachment.java
```diff
@@ -0,0 +1,99 @@
...
+            userName = "user_1_2")
+    @Test
+    public void testGetNoAttachmentSummaryFromTask() {
+        TaskService taskService = taskanaEngine.getTaskService();
+        List<TaskSummary> tasks = taskService.createTaskQuery()
+                .list();
+        assertEquals(20, tasks.size());
+
+        List<AttachmentSummary> attachmentSummaries = tasks.get(0).getAttachmentSummaries();
+        assertEquals(null, attachmentSummaries);
```
TitusLabs: this should not be the case. please, if there are no values in a list, return an empty list and not null

PR #492, opened 2018-07-05
TSK-592 make isValidInDomain work

LarsLeo:


rest/taskana-rest-spring/src/main/java/pro/taskana/rest/resource/ClassificationResource.java
```diff
@@ -31,8 +31,12 @@
     public String custom7;
     public String custom8;
 
-    public void setIsValidInDomain(Boolean validInDomain) {
-        isValidInDomain = validInDomain;
+    public Boolean getIsValidInDomain() {
```
TitusLabs: please use either isValidInDomain (prefered) or getValidInDomain if you must use getter and the mapping is not able to handle is-methods

LarsLeo: The Beanutils.copyProperties() does not seem to work with either of those methods.

TitusLabs: ok, never mind..

web/src/app/administration/services/classifications/classifications.service.ts
```diff
@@ -18,6 +18,7 @@ import { Direction } from 'app/models/sorting';
 @Injectable()
 export class ClassificationsService {
 
+  private url = `${environment.taskanaRestUrl}/v1/classifications/`;
```
MiguelMartinRojas: Why are you setting the url static? we are using HATEOAS

PR #507, opened 2018-07-17
TSK-642 - Remove flickering from classification onSaving a new one

ignaciorecuerda:


web/src/app/administration/classification/details/classification-details.component.ts
```diff
@@ -92,6 +92,7 @@ export class ClassificationDetailsComponent implements OnInit, OnDestroy {
     })
     this.classificationSelectedSubscription = this.classificationsService.getSelectedClassification()
       .subscribe(classificationSelected => {
+        if (this.classification ? this.classification.classificationId === classificationSelected.classificationId : false) { return; }
```
MiguelMartinRojas: Do not use ? : operator withing an if sentence to break the flow, keep it simple, just use short evaluation like this.
if (this.classification && this.classification.classificationId === classificationSelected.classificationId ) { return; }

web/src/app/administration/classification/details/classification-details.component.ts
```diff
@@ -110,10 +111,9 @@ export class ClassificationDetailsComponent implements OnInit, OnDestroy {
         }
         this.fillClassificationInformation(this.selectedClassification ? this.selectedClassification : new ClassificationDefinition())
       }
-
-      if (id && id !== '') {
-        this.selectClassification(id);
-      }
+        if (this.classification ? this.classification.classificationId !== id : true && id && id !== '') {
```
MiguelMartinRojas: Same as above
if(id && id !== '' && this.classification && this.classification.classificationId !== id){ this.selectClassification(id)}

Also extract this.classification ? this.classification.classificationId === classificationSelected.classificationId into a function isSameClassificationId(id) and use it in both if()

PR #512, opened 2018-07-25
TSK-629: Hamonized all job handling.

holgerhagen:


lib/taskana-core/src/main/java/pro/taskana/jobs/JobRunner.java
```diff
@@ -28,35 +38,103 @@ public void registerTransactionProvider(
...
+                runScheduledJob(scheduledJob);
             }
+            jobService.deleteJob(scheduledJob);
+        } catch (Exception e) {
+            e.printStackTrace();
+            // transaction was rolled back -> split job into 2 half sized jobs
+            LOGGER.warn(
+                "Processing of job " + scheduledJob.getJobId() + " failed. Trying to split it up into two pieces...",
+                e);
+            // rescheduleBisectedJob(bulkLog, job);
```
MiguelMartinRojas: We need to implement this mechanism  rescheduleBiselectJob

PR #518, opened 2018-07-28
Tsk 418 Define default schema application using applicationproperties

MiguelMartinRojas:


lib/taskana-core/src/main/java/pro/taskana/configuration/TaskanaEngineConfiguration.java
```diff
@@ -56,14 +56,15 @@
...
-                                                                    // TASKANA.TASKANA_SCHEMA_VERSION
+    // TASKANA_SCHEMA_VERSION
 
     // Taskana properties file
     protected String propertiesFileName = TASKANA_PROPERTIES;
 
     // Taskana datasource configuration
     protected DataSource dataSource;
     protected DbSchemaCreator dbSchemaCreator;
+    protected String defaultSchemaName = "TASKANA";
```
holgerhagen: We should distinguish more precisely between default value and used value. As I understand this is the actually used value. Then we should name it just "schemaName" or "dbSchemaName". 
We can have a constant defining the default value, which is used if no property is set. But that should a different variable.

lib/taskana-core/src/main/java/pro/taskana/configuration/TaskanaEngineConfiguration.java
```diff
@@ -236,7 +240,15 @@ private void initClassificationCategories(Properties props) {
...
-        LOGGER.debug("Configured domains: {}", domains);
+        LOGGER.debug("Configured classification categories : {}", domains);
+    }
+
+    private void initSchemaName(String defaultSchemaName) {
+        if (defaultSchemaName != null && !defaultSchemaName.isEmpty()) {
+            this.setDefaultSchemaName(defaultSchemaName);
+        }
+
+        LOGGER.debug("Working schema: {}", this.getDefaultSchemaName());
```
holgerhagen: I would rather call this "Using schema name {}"

lib/taskana-spring-example/src/main/resources/customdb.properties
```diff
@@ -14,6 +14,7 @@ customdb.datasource.driverClassName=org.h2.Driver
 customdb.datasource.hibernate.dialect=org.hibernate.dialect.H2
 customdb.datasource.username=sa
 customdb.datasource.password=sa
+taskana.defaultSchemaName=TASKANA
```
holgerhagen: This is supposed to be a separate DB. Why do we need this property here? The tables of the custom DB should explicitly be in aonther schema than TASKANA.

rest/taskana-rest-spring-example/src/main/resources/application.properties
```diff
@@ -1,10 +1,22 @@
 logging.level.pro.taskana=DEBUG
 ### logging.level.org.springframework=DEBUG
 ######## Taskana DB #######
+######## h2 configuration ########
 datasource.url=jdbc:h2:mem:taskana;IGNORECASE=TRUE;LOCK_MODE=0;INIT=CREATE SCHEMA IF NOT EXISTS TASKANA
 datasource.driverClassName=org.h2.Driver
 datasource.username=sa
 datasource.password=sa
+taskana.defaultSchemaName=TASKANA
```
holgerhagen: see above. This is not the default value.

PR #527, opened 2018-07-31
TSK-661 - Filtering of distribution target selection does not work

ignaciorecuerda:
Fixed the object of the filter and fixed the test


web/src/app/administration/workbasket/details/distribution-targets/distribution-targets.component.ts
```diff
@@ -148,8 +148,9 @@ export class DistributionTargetsComponent implements OnChanges, OnDestroy {
 		dualListFilter.side === Side.RIGHT ? this.distributionTargetsRight = undefined : this.distributionTargetsLeft = undefined;
 		this.onRequest(dualListFilter.side, false);
 		this.workbasketFilterSubscription = this.workbasketService.getWorkBasketsSummary(true, undefined, undefined, undefined,
-			dualListFilter.filterBy.name, dualListFilter.filterBy.description, undefined, dualListFilter.filterBy.owner,
-			dualListFilter.filterBy.type, undefined, dualListFilter.filterBy.key, undefined, true).subscribe(resultList => {
+      dualListFilter.filterBy.filterParams.name, dualListFilter.filterBy.filterParams.description, undefined,
```
MiguelMartinRojas: There is an error, in filtering component, filter fields shoul remain fullfilled if filter is applied after collapsing the filter.

PR #539, opened 2018-08-08
TSK-644 Make tasks attachments queryable

MiguelMartinRojas:


lib/taskana-core/src/main/java/pro/taskana/mappings/QueryMapper.java
```diff
@@ -112,39 +123,89 @@
...
-        + "CUSTOM_6, CUSTOM_7, CUSTOM_8, CUSTOM_9, CUSTOM_10, CUSTOM_11, CUSTOM_12, CUSTOM_13, CUSTOM_14, CUSTOM_15, CUSTOM_16 FROM Y WHERE FLAG = 1 "
-        + "<if test='!orderBy.isEmpty()'>ORDER BY <foreach item='item' collection='orderBy' separator=',' >${item}</foreach></if> "
+        + "CUSTOM_6, CUSTOM_7, CUSTOM_8, CUSTOM_9, CUSTOM_10, CUSTOM_11, CUSTOM_12, CUSTOM_13, CUSTOM_14, CUSTOM_15, CUSTOM_16"
+        + "<if test=\"joinWithAttachmentsOrderBy\">"
+        + ", ACLASSIFICATION_ID, ACLASSIFICATION_KEY, CHANNEL, REF_VALUE, RECEIVED "
+        + "</if>"
+        + " FROM Y WHERE FLAG = 1 "
+        + "<if test='!orderBy.isEmpty()'>ORDER BY <foreach item='item' collection='orderBy' separator=',' >"
+        + "<choose>"
+        + "<when test=\"item.contains('t.CLASSIFICATION_KEY ASC')\">"
```
holgerhagen: Cannot we prepare this in Java?

lib/taskana-core/src/main/java/pro/taskana/mappings/QueryMapper.java
```diff
@@ -221,6 +282,15 @@
...
+        + "<if test='attachmentChannelIn != null'>AND a.CHANNEL IN(<foreach item='item' collection='attachmentChannelIn' separator=',' >#{item}</foreach>)</if> "
+        + "<if test='attachmentChannelLike != null'>AND (<foreach item='item' collection='attachmentChannelLike' separator=' OR '>UPPER(a.CHANNEL) LIKE #{item}</foreach>)</if> "
+        + "<if test='attachmentReferenceIn != null'>AND a.REF_VALUE IN(<foreach item='item' collection='attachmentReferenceIn' separator=',' >#{item}</foreach>)</if> "
+        + "<if test='attachmentReferenceLike != null'>AND (<foreach item='item' collection='attachmentReferenceLike' separator=' OR '>UPPER(a.REF_VALUE) LIKE #{item}</foreach>)</if> "
+        + "<if test='attachmentReceivedIn !=null'> AND ( <foreach item='item' collection='attachmentReceivedIn' separator=' OR ' > ( <if test='item.begin!=null'> a.RECEIVED &gt;= #{item.begin} </if> <if test='item.begin!=null and item.end!=null'> AND </if><if test='item.end!=null'> a.RECEIVED &lt;=#{item.end} </if>)</foreach>)</if> "
         + "<if test='!orderColumns.isEmpty()'>AND (<foreach item='item' collection='orderColumns' separator='OR ' >${item} IS NOT NULL </foreach>)</if> "
```
holgerhagen: You didn't change anything here, did you? I think this is wrong because I want to the tasks where the value is NULL also

lib/taskana-core/src/main/java/pro/taskana/impl/TaskQueryImpl.java
```diff
@@ -119,6 +128,9 @@
     private List<String> orderBy;
     private List<String> orderColumns;
 
+    private boolean joinWithAttachments = false;
+    private boolean joinWithAttachmentsOrderBy = false;
```
holgerhagen: rather call this "addAttachmentColumnsToSelectClauseForOrdering"?

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
```diff
@@ -731,7 +731,7 @@ private void standardSettings(TaskImpl task, Classification classification,
         }
     }
 
-    List<TaskSummary> augmentTaskSummariesByContainedSummaries(List<TaskSummaryImpl> taskSummaries) {
+    List<TaskSummary> augmentTaskSummariesByContainedSummaries(List<TaskSummaryImpl> taskSummaries, TaskQueryImpl taskQuery) {
```
holgerhagen: I haven't thought about this case. But I think we don't need this. The requirement is rather "Give me a tasks where there is an email attachment". We should wait for a more precise requirment on how to handle multiple attachments in a list (as we already discussed)

lib/taskana-core/src/main/java/pro/taskana/mappings/QueryMapper.java
```diff
@@ -274,48 +344,49 @@
         @Result(property = "custom16", column = "CUSTOM_16")})
     List<TaskSummaryImpl> queryTaskSummaries(TaskQueryImpl taskQuery);
```
holgerhagen: Would it be possible to split this into "queryTaskSummaries" and "queryTasksSummariesOptimizedForDb2"? This is getting too complex. And please add the sample query I've sent as a comment. This helps to understand the structure of the query.

lib/taskana-core/src/main/java/pro/taskana/mappings/QueryMapper.java
MiguelMartinRojas: No I didn't change anything about this "<if test='!orderColumns.isEmpty()'>AND (<foreach item='item' collection='orderColumns' separator='OR ' >${item} IS NOT NULL </foreach>)</if> "
I agree with you, I don't understand why are we removing the null values

lib/taskana-core/src/main/java/pro/taskana/mappings/QueryMapper.java
MiguelMartinRojas: Done

MiguelMartinRojas:
I have applied  all mentioned  changes

PR #543, opened 2018-08-17
TSK-662 Add task classification report to monitor.

MiguelMartinRojas:
- Added task classification report
- Fix tasks status report
- Fix tasks workbasket report
- Remove all harcoded data
- Stylize monitor module according to Admin appearance.


rest/taskana-rest-spring-example/src/test/java/pro/taskana/doc/api/MonitorControllerRestDocumentation.java
```diff
@@ -77,31 +76,53 @@ public void setUp() {
                 fieldWithPath("sumRow.total").description("Total number of tasks"),
                 fieldWithPath("_links.self.href").ignored()
         };
-        
-        countByStateFieldDescriptors = new FieldDescriptor[] {
-                fieldWithPath("[]..state").description("The state the tasks are in"),
-                fieldWithPath("[]..counter").description("Number of tasks in the corresponding state")
+
+        taskStatusReportFieldDescriptors = new FieldDescriptor[] {
```
mustaphazorgati: Seems like a duplicate of lines 65 - 78.

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/MonitorController.java
```diff
@@ -30,53 +33,66 @@
...
             .body(reportAssembler.toResource(
-                taskMonitorService.createTaskStatusReportBuilder().stateIn(states).domainIn(domains).buildReport(),
-                domains, states));
+                taskMonitorService.createClassificationReportBuilder()
+                    .withColumnHeaders(getTaskClassificationTimeInterval())
+                    .buildReport()));
+    }
+
+    private List<TimeIntervalColumnHeader> getTaskClassificationTimeInterval() {
+        List<TimeIntervalColumnHeader> columnHeaders = new ArrayList<>();
```
mustaphazorgati: just for you (don't have to implement this). The stream API is real nice for those operations.

return Stream.concat(
            Stream.of(new TimeIntervalColumnHeader(Integer.MIN_VALUE, -10),
                new TimeIntervalColumnHeader(-10, -5),
                new TimeIntervalColumnHeader(5, 10),
                new TimeIntervalColumnHeader(10, Integer.MAX_VALUE)),
            Stream.of(-4, -3, -2, -1, 0, 1, 2, 3, 4)
                .map(TimeIntervalColumnHeader::new))
            .collect(Collectors.toList());

web/src/app/monitor/services/restConnector/rest-connector.service.ts
```diff
@@ -1,28 +1,106 @@
...
+  parseClassificationHeadersText(header: string, dateFormat = false): string {
+    if (dateFormat) {
+      header = this.dateParseHeaders(header);
+    } else {
+      header = this.defaultParseHeaders(header);
+    }
+    return header;
   }
 
+  defaultParseHeaders(header: string): string {
```
mustaphazorgati: Since we agreed that the ReportData from the rest service provides the display name, i don't like that it is modified. I'd rather prefer a new class which extends `TimeIntervalColumnHeader` where the `displayName` would be overwritten.

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/MonitorController.java
MiguelMartinRojas: Sure I will do it, it's shorter than mine.

web/src/app/monitor/services/restConnector/rest-connector.service.ts
MiguelMartinRojas: Ok, I didn't see displayName from TimeIntervarlColumnHeader, I will extend it and remove the transformation in client side.

PR #545, opened 2018-08-20
TSK-577 - Tasks should only be deleted if all tasks with the same parent business process id can be cleaned up

ignaciorecuerda:
- Add functionality to test if all task with the same parent business process id was completed.
- Add a property to configure this functionality


lib/taskana-core/src/main/java/pro/taskana/jobs/TaskCleanupJob.java
```diff
@@ -61,10 +65,44 @@ public void run() throws TaskanaException {
...
+                .completedWithin(new TimeInterval(null, untilDate))
+                .orderByBusinessProcessId(asc)
+                .list();
+
+        if (allCompletedSameParentBusiness) {
+            Map<String, Long> numberParentTasksShouldHave = new HashMap<>();
+            Map<String, Long> countParentTask = new HashMap<>();
+            for (TaskSummary task : taskList) {
+                numberParentTasksShouldHave.put(task.getParentBusinessProcessId(), taskanaEngineImpl.getTaskService().createTaskQuery().parentBusinessProcessIdIn(task.getParentBusinessProcessId()).count());
+                if (countParentTask.get(task.getParentBusinessProcessId()) != null) {
```
mustaphazorgati: This is way more readable: `countParentTask.merge(task.getParentBusinessProcessId(), 1L, Long::sum);`

lib/taskana-core/src/main/java/pro/taskana/configuration/TaskanaEngineConfiguration.java
```diff
@@ -88,6 +89,7 @@
     private Instant taskCleanupJobFirstRun = Instant.parse("2018-01-01T00:00:00Z");
     private Duration taskCleanupJobRunEvery = Duration.parse("P1D");
     private Duration taskCleanupJobMinimumAge = Duration.parse("P14D");
+    private boolean taskCleanupJobAllCompletedSameParentBusiness = Boolean.parseBoolean("true");
```
mustaphazorgati: why not just `true`?

lib/taskana-core/src/main/java/pro/taskana/jobs/TaskCleanupJob.java
mustaphazorgati: this is way more readable: `countParentTask.merge(task.getParentBusinessProcessId(), 1L, Long::sum);`

lib/taskana-core/src/main/java/pro/taskana/configuration/TaskanaEngineConfiguration.java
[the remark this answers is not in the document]
ignaciorecuerda: Done

PR #553, opened 2018-09-03
TSK-674: now displaying workbasked when opening a task using a deep link

mustaphazorgati:


web/tslint.json
```diff
@@ -47,10 +47,6 @@
 		"no-empty": false,
 		"no-empty-interface": true,
 		"no-eval": true,
-		"no-inferrable-types": [
```
MiguelMartinRojas: Why?

PR #558, opened 2018-09-11
TSK-701

LarsLeo:
is this oneliner enough, or should the text be dependant on what the user did before?


web/src/app/workplace/tasklist/tasklist.component.html
```diff
@@ -6,7 +6,7 @@
   <div *ngIf="!requestInProgress">
     <ul #taskList id="task-list-container" class="list-group">
       <li class="list-group-item" *ngIf="tasks === undefined || tasks.length === 0" type="text">
-        <b>This Workbasket has no Tasks</b>
```
MiguelMartinRojas: Text is ok, but add an icon like workbasket list

mustaphazorgati: Not only that. It has to be a different container since the "This workbasket has no Tasks" message should still be there :)

PR #559, opened 2018-09-11
TSK-679 - Delete all workbasket access items for a user

ignaciorecuerda:


rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketAccessItemController.java
```diff
@@ -86,6 +100,40 @@
...
+            throw new InvalidArgumentException(
+                accessId + " corresponding to a group, not a user. You just can remove access items for a user");
+        }
+
+        return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
+    }
+
+    private boolean isGroup(String accessId) {
+        if ((ldapClient.useLdap() && accessId.contains(ldapClient.getGroupSearchBase())) || (!ldapClient.useLdap()
+            && accessId.contains("ou=groups"))) {
```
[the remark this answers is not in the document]
ignaciorecuerda: You are right. Thank you, Holger

rest/taskana-rest-spring-example/src/test/java/pro/taskana/rest/WorkbasketAccessItemControllerIntTest.java
```diff
@@ -101,6 +104,30 @@ public void testGetSecondPageSortedByWorkbasketKey() {
...
+        ResponseEntity<Void> response = template.exchange(
+                url + port + parameters, HttpMethod.DELETE, request,
+                new ParameterizedTypeReference<Void>() {
+                });
+        assertNull(response.getBody());
+        assertEquals(HttpStatus.NO_CONTENT, response.getStatusCode());
+    }
+
+    @Test
+    public void testRemoveWorkbasketAccessItemsOfGroup() {
```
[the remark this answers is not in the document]
ignaciorecuerda: Done

rest/taskana-rest-spring-example/src/main/asciidoc/rest-api.adoc
```diff
@@ -343,6 +343,14 @@ include::../../../{snippets}/DeleteClassificationDocTest/http-request.adoc[]
 
 include::../../../{snippets}/DeleteClassificationDocTest/http-response.adoc[]
 
+=== GET all workbasketAccessItems
```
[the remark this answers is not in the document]
ignaciorecuerda: You are right. I put the new resource on the correct position. On the other hand, Delete classification is correct, so I didn't change their position

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketAccessItemController.java
```diff
@@ -86,6 +100,36 @@
...
     }
 
+    /**
+     * This DELETE method delete all workbasketAccessItems that correspond the given accessId.
+     *
+     * @param accessId which need remove his workbasketAccessItems.
+     * @return ResponseEntity if the user is not authorized.
+     * @throws NotAuthorizedException if the user is not authorized.
+     */
+    @DeleteMapping("{accessId}")
```
holgerhagen: I understand that this is not scope of this PR, bur can we change the endpoint to "/v1/workbasket-access-items"? We should try to be consistent with out naming and the resource we are getting back is a WorkbasketAccessItem, isn't it?

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketAccessItemController.java
```diff
@@ -86,6 +100,36 @@
...
+    public ResponseEntity<Void> removeWorkbasketAccessItems(@PathVariable("accessId") String accessId)
```
holgerhagen: And one more comment to consistency: the path variable should be a uinque id. "accessId" is not identifying a single WorkbasketAccessItem. It can identify a list of items. Therefore we should not use accessId as a PathVariable but rather a Parameter. The call would than look like:
"DELETE /v1/workbasket-access-items?access-id=%%USERID%%"

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketAccessItemController.java
```diff
@@ -86,6 +100,36 @@
...
+            }
+        } else {
+            throw new InvalidArgumentException(
+                accessId + " corresponding to a group, not a user. You just can remove access items for a user");
+        }
+
+        return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
+    }
+
+    private boolean isGroup(String accessId) {
```
holgerhagen: That is good. But we can improve this still a bit and move the logix to the LdapClient. Then we would encapsulate all Ldap configuration there and the REST controller would not need to know about it. What do you think?

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketAccessItemController.java
ignaciorecuerda: It's true. I should be careful with the consistency of the names. 
Thank you!

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketAccessItemController.java
ignaciorecuerda: Done!

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketAccessItemController.java
ignaciorecuerda: You are right! It's better than all about ldap is on ldap client.
Thank you for your advice

PR #560, opened 2018-09-17
TSK-670 Remove IS NOT NULL condition for queryMapper and fix failing tests

MiguelMartinRojas:


lib/taskana-core/src/main/java/pro/taskana/mappings/QueryMapper.java
```diff
@@ -131,7 +131,7 @@ static String fixColumnNames() {
         + "<if test='attachmentReferenceIn != null'>AND a.REF_VALUE IN(<foreach item='item' collection='attachmentReferenceIn' separator=',' >#{item}</foreach>)</if> "
         + "<if test='attachmentReferenceLike != null'>AND (<foreach item='item' collection='attachmentReferenceLike' separator=' OR '>UPPER(a.REF_VALUE) LIKE #{item}</foreach>)</if> "
         + "<if test='attachmentReceivedIn !=null'> AND ( <foreach item='item' collection='attachmentReceivedIn' separator=' OR ' > ( <if test='item.begin!=null'> a.RECEIVED &gt;= #{item.begin} </if> <if test='item.begin!=null and item.end!=null'> AND </if><if test='item.end!=null'> a.RECEIVED &lt;=#{item.end} </if>)</foreach>)</if> "
-        + "<if test='!orderColumns.isEmpty()'>AND (<foreach item='item' collection='orderColumns' separator='OR ' >${item} IS NOT NULL </foreach>)</if> "
+
```
holgerhagen: we do not need this empty line, do we?

lib/taskana-core/src/test/java/acceptance/task/QueryTaskWithAttachment.java
```diff
@@ -74,24 +74,24 @@ public void testIfNewTaskHasEmptyAttachmentList() {
...
+            .customAttributeIn("1", "custom1")
+            .list();
         assertEquals(1, tasks.size());
 
         Task originalTask = taskService.getTask("TKI:000000000000000000000000000000000000");
         Attachment originalAttachment = originalTask.getAttachments().get(0);
 
         // Test if it's the Summary of the Original Attachment
-        AttachmentSummary queryAttachmentSummary = tasks.get(0).getAttachmentSummaries().get(0);
+        AttachmentSummary queryAttachmentSummary = tasks.get(0).getAttachmentSummaries().get(1);
```
holgerhagen: this looks not safe to me. Maybe we should explicitly compare the attechments with the same id or some other match code? 
Why is the order different in the two case and element 0 in one call is element 1 in the other?

MiguelMartinRojas: Distinct can change the order of the query.

PR #563, opened 2018-09-18
TSK-681 Create workbasket cleanup job

MiguelMartinRojas:


lib/taskana-core/src/main/java/pro/taskana/impl/WorkbasketServiceImpl.java
```diff
@@ -779,19 +780,80 @@ private void markWorkbasketForDeletion(String workbasketId)
...
+        taskanaEngine.checkRoleMembership(TaskanaRole.BUSINESS_ADMIN, TaskanaRole.ADMIN);
+
+        try {
+            taskanaEngine.openConnection();
+            if (workbasketsIds == null || workbasketsIds.isEmpty()) {
+                throw new InvalidArgumentException("List of WorkbasketIds must not be null.");
+            }
+
+            List<String> existingWorkbasketIds = workbasketMapper.findExistingWorkbaskets(workbasketsIds);
+            BulkOperationResults<String, TaskanaException> bulkLog = cleanNonExistingWorkbasketExists(
```
holgerhagen: I don't understand what this is good for. Just to verify that the workbasket (id) exists? Since we run into deleteWorkbasket later on, this check will be done implicitly, won't it? 
And I would expect to have the errors/exceptions from deleteWorkbasket in the bulklog. Is there a reason for doing it this way?

lib/taskana-core/src/main/java/pro/taskana/jobs/WorkbasketCleanupJob.java
```diff
@@ -0,0 +1,172 @@
...
+            scheduleNextCleanupJob();
+        }
+    }
+
+    private List<String> getWorkbasketsMarkedForDeletion() throws InvalidArgumentException {
+        List<String> workbasketList = taskanaEngineImpl.getWorkbasketService()
+            .createWorkbasketQuery()
+            .markedForDeletion(true)
+            .listValues("ID", BaseQuery.SortDirection.ASCENDING);
+        workbasketList = excludeWorkbasketWithPendingTasks(workbasketList);
```
holgerhagen: This is a good idea. But wouldn't this be better placed in the deleteWorkbaskets() operation instead of the cleanup job?

lib/taskana-core/src/main/java/pro/taskana/mappings/WorkbasketMapper.java
```diff
@@ -150,6 +152,12 @@
         @Result(property = "orgLevel4", column = "ORG_LEVEL_4")})
     List<WorkbasketSummaryImpl> findAll();
 
+    @Select("<script>SELECT ID FROM WORKBASKET "
+        + "WHERE ID IN( <foreach item='item' collection='workbasketIds' separator=',' >#{item}</foreach> ) "
+        + "<if test=\"_databaseId == 'db2'\">with UR </if> "
+        + "</script>")
+    List<String> findExistingWorkbaskets(@Param("workbasketIds") List<String> workbasketIds);
```
holgerhagen: This might not be required. If it is, maybe it can be replaced by a "standard workbasket query"?

lib/taskana-core/src/main/java/pro/taskana/impl/WorkbasketServiceImpl.java
MiguelMartinRojas: Ok, instead of returning a bulklog with non existing workbaskets, I will filter the list excludingWorkbasketWithPendingTasks which I think it is more usefull.

PR #564, opened 2018-09-18
TSK-24 create new task

LarsLeo:
The user is now able to create a task in a selected workbasket in the workplace.
I also moved the classification services to app/services since they need to be used in workplace and admin.


web/src/app/services/classifications/classifications.service.ts
```diff
@@ -8,7 +8,7 @@ import { Classification } from 'app/models/classification';
 import { ClassificationDefinition } from 'app/models/classification-definition';
```
MiguelMartinRojas: General UI issues:
- Remove add button if there is no workbasket selected
- You need to validate the form because if you try to save without entering any field the error message is returning an error but no field is displayed as required or needed
- I would suggest place required fields at the beginning of the form (better readibility)
- I am not sure if classificaiton dropdown is bootstrap's default if not, use it (the same as classification screen)
- I think there is no need for showing non-editable fields like creation date ... since you are creating a task and the can't be modified
- I would suggest to add a yellow badge close to the title, like creation of a workbasket(Creating new workbasket)
- Title doesn't change and there is no field wich makes it change
- All references values seems to be mandatory, please validate the form before saving.

web/src/app/workplace/services/task.service.ts
```diff
@@ -82,6 +91,14 @@ export class TaskService {
     return this.httpClient.delete<Task>(`${this.url}/${task.taskId}`);
   }
 
+  createTask(task: Task): Observable<Task> {
+    return this.httpClient.post<Task>(`${this.url}`, task);
+  }
+
+  newTask(workbasket: Workbasket): Observable<Task> {
```
MiguelMartinRojas: Remove newTask it is not been used

web/src/app/workplace/taskdetails/taskdetails.component.ts
```diff
@@ -101,7 +139,7 @@ export class TaskdetailsComponent implements OnInit, OnDestroy {
   backClicked(): void {
     this.task = undefined;
     this.taskService.selectTask(this.task);
-    this.router.navigate(['./'], { relativeTo: this.route.parent });
+    this.router.navigate(['./'], {relativeTo: this.route.parent});
```
MiguelMartinRojas: If you go directly to -> http://localhost:4200/workplace/tasks/(detail:taskdetail/new-task)
deeplinking doesn't work but new-task screen is been loaded, allow deepliking or avoid an user navigate directly to this route

web/src/app/workplace/taskdetails/taskdetails.component.ts
```diff
@@ -1,80 +1,119 @@
...
       this.taskService.publishUpdatedTask(task);
       this.alertService.triggerAlert(new AlertModel(AlertType.SUCCESS, 'Update successful!'))
-    }, err => {this.alertService.triggerAlert(new AlertModel(AlertType.DANGER, 'Update not successful!'))});
+    }, err => {
+      this.alertService.triggerAlert(new AlertModel(AlertType.DANGER, 'Update not successful!'))
+    });
+  }
+
+  createTask() {
+    this.addDateToTask();
```
MiguelMartinRojas: You are missing to activate general spinnner this.requestInProgressService.setRequestInProgress(true);

web/src/app/workplace/tasklist/tasklist.component.scss
```diff
@@ -16,10 +16,9 @@
 }
 
 ul {
-  max-height: 90vh;
+  max-height: 88.5vh;
```
MiguelMartinRojas: why? if you are playing with vh is probably because you need to have a look to site.scss
taskana-workbasket-information,taskana-task-details, taskana-workbasket-access-items, taskana-workbaskets-distribution-targets, taskana-monitor-tasks,
taskana-monitor-workbaskets, taskana-monitor-classification-tasks, taskana-access-items-management, taskana-classification-details,taskana-workbasket-details {
    & .panel{
        border: none;
        box-shadow: none;
        margin-bottom: 0px;
        &> .panel-body {
            height: calc(100vh - 108px);
            max-height: calc(100vh - 108px);
            overflow-y: auto;
        }
    }
}

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/TaskController.java
```diff
@@ -90,7 +90,7 @@
     @GetMapping
     @Transactional(readOnly = true, rollbackFor = Exception.class)
     public ResponseEntity<PagedResources<TaskSummaryResource>> getTasks(
-        @RequestParam MultiValueMap<String, String> params) throws InvalidArgumentException {
+            @RequestParam MultiValueMap<String, String> params) throws InvalidArgumentException {
```
MiguelMartinRojas: We need to use the same linting I can't identify what do you have changed in rest part.

web/src/app/workplace/taskdetails/taskdetails.component.ts
LarsLeo: I am doing that in onSave()

web/src/app/workplace/services/task.service.ts
LarsLeo: done

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/TaskController.java
LarsLeo: yes, nothing should have changed here, this actually should be the master version.

web/src/app/services/classifications/classifications.service.ts
LarsLeo: * done
* own ticket for validation, needs to be done not only when adding tasks
* done
* done
* done
* done
* done
* see 2

mustaphazorgati:
You got an IntelliJ setting wrong. Please configure as attached picture and make sure that you don't have a diff with wrong indentation. Thanks!
image

PR #572, opened 2018-10-02
TSK-714 - Improve all workbasket access items for a user screen

ignaciorecuerda:
- Create new endpoint to get the groups associated to an access id (validating the access id)
- Show in the popup the group associated to access id (only for a user)
- TypeAhead is red when the access id is invalid.
- The delete button is on bottom right position (just activated for the user)
- Disable delete button for groups
- Remove the dots of the lists

### DIFF
--- a/rest/taskana-rest-spring/src/main/java/pro/taskana/rest/AccessIdController.java
+++ b/rest/taskana-rest-spring/src/main/java/pro/taskana/rest/AccessIdController.java
@@ -35,28 +35,40 @@ public class AccessIdController {
 
     @GetMapping
     public ResponseEntity<List<AccessIdResource>> validateAccessIds(
-        @RequestParam String searchFor, @RequestParam(required = false) boolean searchInGroups) throws InvalidArgumentException {
+        @RequestParam("search-for") String searchFor) throws InvalidArgumentException {
         if (searchFor.length() < ldapClient.getMinSearchForLength()) {
-            throw new InvalidArgumentException("searchFor string '" + searchFor + "' is too short. Minimum searchFor length = "
-                + ldapClient.getMinSearchForLength());
+            throw new InvalidArgumentException(
+                "searchFor string '" + searchFor + "' is too short. Minimum searchFor length = "
+                    + ldapClient.getMinSearchForLength());
         }
         if (ldapClient.useLdap()) {
             List<AccessIdResource> accessIdUsers = ldapClient.searchUsersAndGroups(searchFor);
-            if (searchInGroups) {
-                List<AccessIdResource> accessIdGroups = ldapClient.searchGroupsofUsersIsMember(searchFor);
-                accessIdUsers.addAll(accessIdGroups);
-            }
             return new ResponseEntity<>(accessIdUsers, HttpStatus.OK);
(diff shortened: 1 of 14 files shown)


rest/taskana-rest-spring/src/main/java/pro/taskana/rest/AccessIdController.java
```diff
@@ -35,28 +35,37 @@
...
-            }
+            return new ResponseEntity<>(
+                ldapCache.findMatchingAccessId(searchFor, ldapClient.getMaxNumberOfReturnedAccessIds()),
+                HttpStatus.OK);
+        } else {
+            return new ResponseEntity<>(new ArrayList<>(), HttpStatus.NOT_FOUND);
+        }
+    }
+
+    @GetMapping(path = "/get-groups")
```
holgerhagen: We should always use an entity for the path. In this case 'groups'. GET is the HTTP method.

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/AccessIdController.java
```diff
@@ -66,4 +75,18 @@ public static void setLdapCache(LdapCache cache) {
         ldapCache = cache;
     }
 
+    private List<AccessIdResource> validateAccessId(String accessId) throws InvalidArgumentException {
```
holgerhagen: does this really validate something? It actually searches all users and groups matching the String, doesn't it? I prefer to have more meaningful names. The code itself is good.

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/AccessIdController.java
```diff
@@ -66,4 +75,18 @@ public static void setLdapCache(LdapCache cache) {
...
+        List<AccessIdResource> accessIdUsers;
+        if (ldapClient.useLdap()) {
+            accessIdUsers = ldapClient.searchUsersAndGroups(accessId);
+        } else {
+            accessIdUsers = ldapCache.validateAccessId(accessId);
+        }
+
+        if (accessIdUsers.size() != 1) {
```
holgerhagen: ... but why do we check for the size 1 if the return type is a list?

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/AccessIdController.java
ignaciorecuerda: Done

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/AccessIdController.java
ignaciorecuerda: This method just tests if the accessId is valid. Return a boolean

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/AccessIdController.java
ignaciorecuerda: I check if the size is just 1 to validate the accessId. Now is more logical because the method just check the validation

PR #573, opened 2018-10-09
TSK-729 Problems importing classifications

ignaciorecuerda:
The problem was that the classificationId was losing because it doesn't know which is the parent id. I resolve this problem creating a map to mapper the old classificationId and the new classificationId.


holgerhagen:
Closed since this is handled in another PR.

PR #577, opened 2018-10-16
TSK-728 - Update bootstrap version

ignaciorecuerda:


web/src/assets/custom.scss
```diff
@@ -0,0 +1,5 @@
+.dropdown-menu.show {
```
MiguelMartinRojas: rename custom to .... bootstrap-3-backward-compatibility

PR #578, opened 2018-10-16
TSK-681 Workbasket cleanup job.

MiguelMartinRojas:


lib/taskana-core/src/test/java/acceptance/jobs/WorkbasketCleanupJobAccTest.java
```diff
@@ -0,0 +1,111 @@
...
+        assertEquals(numTasksNotCompletedInWorkbasket, 0);
+        try {
+            workbasketService.deleteWorkbasket(workbaskets.get(1).getId());
+            workbasketService.deleteWorkbasket(workbaskets.get(0).getId());
+        } catch (NotAuthorizedException | WorkbasketNotFoundException | WorkbasketInUseException | InvalidArgumentException e) {
+            // workbaskets.get(0) contains non-completed tasks and can´t be marked for deletion.
+        }
+
+        WorkbasketCleanupJob job = new WorkbasketCleanupJob(taskanaEngine, null, null);
+        job.run();
```
holgerhagen: Please give it a try commenting out this line. It should still work because an empty workbasket is deleted right away.

lib/taskana-core/src/main/java/pro/taskana/impl/WorkbasketServiceImpl.java
```diff
@@ -779,19 +783,84 @@ private void markWorkbasketForDeletion(String workbasketId)
...
+            if (workbasketsIds == null || workbasketsIds.isEmpty()) {
+                throw new InvalidArgumentException("List of WorkbasketIds must not be null.");
+            }
+            WorkbasketQuery workbasketQuery = createWorkbasketQuery();
+
+            List<String> existingWorkbasketsIds = workbasketQuery.idIn(workbasketsIds.stream().toArray(String[]::new))
+                .listValues("ID",
+                    BaseQuery.SortDirection.DESCENDING);
+
+            BulkOperationResults<String, TaskanaException> bulkLog = excludeWorkbasketWithPendingTasks(workbasketsIds);
```
holgerhagen: Please just iterate over deleteWorkbasket, collecting the exceptions.

lib/taskana-core/src/main/java/pro/taskana/impl/WorkbasketServiceImpl.java
```diff
@@ -779,19 +782,67 @@ private void markWorkbasketForDeletion(String workbasketId)
...
+        taskanaEngine.checkRoleMembership(TaskanaRole.BUSINESS_ADMIN, TaskanaRole.ADMIN);
+
+        try {
+            taskanaEngine.openConnection();
+            if (workbasketsIds == null || workbasketsIds.isEmpty()) {
+                throw new InvalidArgumentException("List of WorkbasketIds must not be null.");
+            }
+            WorkbasketQuery workbasketQuery = createWorkbasketQuery();
+
+            List<String> existingWorkbasketsIds = workbasketQuery.idIn(workbasketsIds.stream().toArray(String[]::new))
```
holgerhagen: Actually we don't even need this call. deleteWorkbasket will find out if the workbasket exists. If it doesn't the error will be stored in the bulk log. 
If we do this call, we will read the entire workbasket table even if we just want to delete 2 workbaskets.

lib/taskana-core/src/test/java/acceptance/jobs/WorkbasketCleanupJobAccTest.java
holgerhagen: Did you change something regarding the acceptance tests? I don't see it actually...

lib/taskana-core/src/main/java/pro/taskana/impl/WorkbasketServiceImpl.java
MiguelMartinRojas: Yes, you are right. I will remove it

lib/taskana-core/src/test/java/acceptance/jobs/WorkbasketCleanupJobAccTest.java
MiguelMartinRojas: I am going to redo the tests because now they looks weird.

MiguelMartinRojas: I have redo the tests because they looks weird.

lib/taskana-core/src/main/java/pro/taskana/impl/WorkbasketServiceImpl.java
MiguelMartinRojas: You are right, Now it's deleted

PR #588, opened 2018-10-25
TSK-732: Refactored BaseQuery.listValues method to use Enum instead of String

JulianSchallenmueller:


lib/taskana-core/src/test/java/acceptance/classification/QueryClassificationAccTest.java
```diff
@@ -39,30 +40,30 @@ public QueryClassificationAccTest() {
     public void testQueryClassificationValuesForColumnName() {
         ClassificationService classificationService = taskanaEngine.getClassificationService();
         List<String> columnValueList = classificationService.createClassificationQuery()
-            .listValues("NAME", null);
+            .listValues(ClassificationQueryColumnName.NAME, null);
```
TitusLabs: a static import could be nice for these use cases..

lib/taskana-core/src/test/java/acceptance/workbasket/QueryWorkbasketAccTest.java
```diff
@@ -94,14 +95,14 @@ public void testQueryAllForAdminMultipleTimes() {
     public void testQueryWorkbasketValuesForColumnName() {
         WorkbasketService workbasketService = taskanaEngine.getWorkbasketService();
         List<String> columnValueList = workbasketService.createWorkbasketQuery()
-            .listValues("NAME", null);
+            .listValues(WorkbasketQueryColumnName.NAME, SortDirection.ASCENDING);
```
TitusLabs: may be keep the default by using null

lib/taskana-core/src/test/java/acceptance/classification/QueryClassificationAccTest.java
JulianSchallenmueller: I have added the static imports now.

lib/taskana-core/src/test/java/acceptance/workbasket/QueryWorkbasketAccTest.java
JulianSchallenmueller: Yes, this was a little oversight by me, I had it changed at an earlier stage, but now it is back how it was.

lib/taskana-core/src/test/java/acceptance/classification/QueryClassificationAccTest.java
JulianSchallenmueller: I just noticed I have forgotten to add it in the AccessItemQuery, I'm at it now.

lib/taskana-core/src/main/java/pro/taskana/TaskQueryColumnName.java
```diff
@@ -0,0 +1,70 @@
...
+    CUSTOM_8("t.custom_8"),
+    CUSTOM_9("t.custom_9"),
+    CUSTOM_10("t.custom_10"),
+    CUSTOM_11("t.custom_11"),
+    CUSTOM_12("t.custom_12"),
+    CUSTOM_13("t.custom_13"),
+    CUSTOM_14("t.custom_14"),
+    CUSTOM_15("t.custom_15"),
+    CUSTOM_16("t.custom_16"),
+    ACLASSIFICATION_KEY("a.classification_key"),
```
TitusLabs: how about a underscore after the A for attachment?

JulianSchallenmueller: I have added an 'A_' for all attributes from Attachment, I think that would be a more consistent naming.

PR #598, opened 2018-11-12
TSK-726 Add HistoryPluginLoaderTests with profile option

MiguelMartinRojas:


lib/taskana-spring-example/src/test/java/pro/taskana/TaskanaTransactionTest.java
```diff
@@ -269,6 +265,12 @@ public void testWorkbasketCleanupJobTransaction() {
             e.printStackTrace();
         }
     }
+    
+    @Test
+    public void testHistoryEventProducerIsNotEnabled() {
```
holgerhagen: Can we move this to a dedicated test class? This has nothing to do with transactions, does it?

Can we also change TaskanaTransactionTest to TaskanaTransactionIntTest? This is rather an INT test, isn't it?

PR #599, opened 2018-11-12
TSK-740 Remove glyphy icon library due to deprecation in bootstrap 4

MiguelMartinRojas:


web/package.json
```diff
@@ -49,7 +50,7 @@
     "hoek": "5.0.3",
     "jasmine-core": "2.9.1",
     "jasmine-spec-reporter": "4.2.1",
-    "karma": "2.0.4",
+    "karma": "^3.1.1",
```
holgerhagen: this is a "min" dependency, right? wouldn't a fixed dependency be better?

MiguelMartinRojas: Sure, they are added automatically with ^ on updating. I wil change it.

PR #600, opened 2018-11-12
TSK-647 - [UI] Add pagination or infite scrolling to distribution targets lists in order to avoid loading the whole workbasket list.

ignaciorecuerda:


web/src/app/administration/workbasket/details/distribution-targets/distribution-targets.component.ts
```diff
@@ -100,19 +107,88 @@ export class DistributionTargetsComponent implements OnChanges, OnDestroy {
...
 			});
 
-	}
+    this.orientationSubscription = this.orientationService.getOrientation().subscribe((orientation: Orientation) => {
+      this.calculateNumberItemsList();
+      this.getWorkbaskets();
+    });
+  }
+
+  calculateNumberItemsList() {
```
MiguelMartinRojas: What do you think about moving this metod into orientationService, using parameters and be as most minimal as you can.

web/src/app/administration/workbasket/details/distribution-targets/distribution-targets.component.ts
```diff
@@ -100,19 +107,88 @@ export class DistributionTargetsComponent implements OnChanges, OnDestroy {
...
+          }
+          this.onRequest(undefined, true);
+          if (!this.pageLeft || !this.pageRight) {
+            this.calculatePages(distributionTargetsAvailable);
+          }
+        });
+  }
+
+  calculatePages(distributionTargetsAvailable) {
+    this.pageLeft = new Page(this.cards,
```
MiguelMartinRojas: We are introducing a lot of complexity into this component, we need to rethink the approach, try to implement infinite scrolling instead of pagination.

PR #605, opened 2018-11-22
Fixed bluemix deployment and modified script for allowing to force ployment

MiguelMartinRojas:


ci/copy-rest-spring.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e # fail fast
 
-if [[ ! "$1" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then 
+if [[ ! "$1" =~ ^v[0-9]+\.[0-9]+\.[0-9] ]]; then
```
mustaphazorgati: be more specific: ^v[0-9]+\.[0-9]+\.[0-9]+(-SNAPSHOT)?$
And that "+" is supposed to stay there ;)

.travis.yml
```diff
@@ -46,27 +49,32 @@ jobs:
...
-    - stage: "Release / Deploy"
+    - stage: Release
       script:
         - ci/release.sh lib/taskana-core $TRAVIS_TAG
           && ci/release.sh lib/taskana-spring $TRAVIS_TAG
           && ci/release.sh lib/taskana-cdi $TRAVIS_TAG
           && ci/release.sh web/ $TRAVIS_TAG
           && ci/release.sh rest/taskana-rest-spring $TRAVIS_TAG
       if: repo = env(DEPLOY_REPO) AND (tag =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ OR branch = master) AND type != pull_request
+    - stage: AfterRelease
```
mustaphazorgati: AfterRelease is not saying anything.
"Update POMs" sounds a lot better :)

.travis.yml
```diff
@@ -14,10 +14,13 @@ cache:
...
   - Build
   - Test
-  - "Release / Deploy"
+  - Release
+  - AfterRelease
+  - Deploy
 
 env:
   global:
+    - FORCE_DEPLOY_TAG=null # Change to taskana version like: v1.0.5-SNAPSHOT to force the deployment.
```
mustaphazorgati: We actually have an issue here: This only works if the given version is saved in the travis cache or is the current version.
We can either leave it like this and have this as a convention or read the current version out of the pom file (either via bash or mvn itself)

I prefer the ladder since it creates less mistakes.

Code to extract version: 

<code>VERSION=\`mvn -q -Dexec.executable="echo" -Dexec.args='${project.version}' --non-recursive org.codehaus.mojo:exec-maven-plugin:1.6.0:exec -f ../lib\`</code>

(only working from ci folder)

.travis.yml
```diff
@@ -46,27 +49,32 @@ jobs:
...
           && ci/release.sh rest/taskana-rest-spring $TRAVIS_TAG
       if: repo = env(DEPLOY_REPO) AND (tag =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ OR branch = master) AND type != pull_request
+    - stage: AfterRelease
+      script: 
+        - ci/change_version.sh -i -m "lib/ rest/" -swarm lib/taskana-cdi/src/test/java/pro/taskana/TaskanaProducersTest.java
+          && ci/commitPoms.sh lib/taskana-cdi/src/test/java/pro/taskana/TaskanaProducersTest.java
+      if: repo = env(DEPLOY_REPO) AND tag =~ ^v[0-9]+\.[0-9]+\.[0-9]+$
+    - stage: Deploy
+      script: 
+        - if [[ $FORCE_DEPLOY_TAG =~ ^v[0-9]+\.[0-9]+\.[0-9] ]];
```
mustaphazorgati: With the changes from the first commit i'd modify the `copy-rest-spring.sh` file to use two parameters. The tag and force deploy.

mustaphazorgati:
These changes also have a differnet problem: Due to the introduction of the 2 new stages deploy / release is now executed in sequence which results in a longer build time.

I'd unify those 3 stages back to one stage, so that those 3 jobs can be run in parralel.

.travis.yml
```diff
@@ -46,27 +49,32 @@ jobs:
...
+      deploy:
+        provider: cloudfoundry
+        username: $BLUEMIX_ACCOUNT
+        password: $BLUEMIX_ACCOUNT_PASSWORD
+        api: https://api.ng.bluemix.net
+        organization: "NovaTec Consulting GmbH"
+        space: Taskana
+        on:
+          all_branches: true
+      if: (repo = env(DEPLOY_REPO) AND tag =~ ^v[0-9]+\.[0-9]+\.[0-9]+$) OR (env(FORCE_DEPLOY_TAG) =~ ^v[0-9]+\.[0-9]+\.[0-9])
```
mustaphazorgati: It is not possible to have a tag during a PR. That's why I didn't check for it here.
Is it possible to custom trigger a PR build? If yes, then please catch that - We never ever want to deploy a PR build :)

.travis.yml
MiguelMartinRojas: Ok I will use it, this way we won't need to specify the version for forcing deployment

.travis.yml
MiguelMartinRojas: That's depend on the purpose of the production web application if  it is really a production application or it is just a demo application, If it is a demo I don't see why we don't want to deploy PR, if is for production  purpose we should only deply master.

MiguelMartinRojas:
> These changes also have a differnet problem: Due to the introduction of the 2 new stages deploy / release is now executed in sequence which results in a longer build time.
> 
> I'd unify those 3 stages back to one stage, so that those 3 jobs can be run in parralel.
> 
> What do you think @MiguelMartinRojas ?

There is no other way to parallelize task in travis? I understand that call them with the same name will run then in parallel but for readibility they should be named different.

.travis.yml
MiguelMartinRojas: Ok, makes sense

.travis.yml
mustaphazorgati: @holgerhagen I think Miguel and I are not in a position to decide this. What's your call?

mustaphazorgati:
> There is no other way to parallelize task in travis? I understand that call them with the same name will run then in parallel but for readibility they should be named different.

Unfortunately those are the limitations of travis. Its build stages are fresh out of beta. Luckily this is already a known issue and a well wanted feature from the community https://github.com/travis-ci/travis-ci/issues/5898 .

There are some workarounds which are acceptable. One of them is working with the enviornment variables (which are displayed in the view). We have to be careful with the cache tho. Environment variables determine which cache is used. See https://docs.travis-ci.com/user/caching/#caches-and-build-matrices

.travis.yml
holgerhagen: A PR needs to be integrated before it makes sense to deploy it, or am I getting something wrong here? We would only deploy the master branch which would explude the PRs, wouldn't it?

mustaphazorgati: That's all I wanted to hear. - Let's make the restriction that the branch hast to be master and not a pr

mustaphazorgati:
Nevermind... Look what I just found: https://docs.travis-ci.com/user/customizing-the-build/#naming-jobs-within-matrices

PR #606, opened 2018-11-22
TSK-739 - Fix task list filtering

ignaciorecuerda:
- The filter of state is now a dropdown
- The search by Name and owner is with "LIKE"

### DIFF
--- a/web/src/app/shared/filter/filter.component.html
+++ b/web/src/app/shared/filter/filter.component.html
@@ -65,13 +65,23 @@
       </button>
     </div>
     <div class="row">
-      <div class="col-xs-4 col-xs-offset-2">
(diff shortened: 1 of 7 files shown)


web/src/app/shared/filter/filter.component.html
```diff
@@ -64,15 +64,25 @@
         <span class="material-icons md-20 blue">clear</span>
       </button>
     </div>
-    <div class="row">
-      <div class="col-xs-4 col-xs-offset-2">
-        <input type="text" [(ngModel)]="filter.filterParams.state" (keyup.enter)="search()" class="form-control" id="display-owner-state"
-          placeholder="Filter status">
+    <div class="row" style="margin-top: 1px;">
```
MiguelMartinRojas: DO NOT USE INLINE STYLING

ignaciorecuerda: I remove all my inline styling. I change to selector style

PR #611, opened 2018-11-29
TSK-617 - Show build version in UI

ignaciorecuerda:


rest/taskana-rest-spring/src/main/java/pro/taskana/rest/TaskanaEngineController.java
```diff
@@ -66,7 +70,6 @@
      */
     @GetMapping(path = "/v1/version", produces = {MediaType.APPLICATION_JSON_VALUE})
     public ResponseEntity<VersionResource> currentVersion() {
-        String version = TaskanaEngineController.class.getPackage().getImplementationVersion();
         if (version == null) {
             version = "1.0-DEFAULT";
```
holgerhagen: in what cases don't we have a version set? Maybe we can set a more meaningful message for such case? (like "local build" for example?)

ignaciorecuerda: It's a good idea

ignaciorecuerda:
Regarding default property value, I set a "Local build" to the default value (for example if the property doesn't exist)

PR #612, opened 2018-11-29
TSK-754: PrettyPrint for JSON added

JulianSchallenmueller:


rest/taskana-rest-spring/src/main/java/pro/taskana/rest/RestConfiguration.java
```diff
@@ -32,11 +39,14 @@
 @Configuration
 @ComponentScan
 @EnableTransactionManagement
-public class RestConfiguration {
+public class RestConfiguration extends WebMvcConfigurationSupport {
```
MiguelMartinRojas: extending WebMvcConfigurationSupport  doesn't allow me to run taskana-rest-spring-example correctly

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/RestConfiguration.java
```diff
@@ -88,4 +98,10 @@ public Jackson2ObjectMapperBuilder jacksonBuilder(HandlerInstantiator handlerIns
     public HandlerInstantiator handlerInstantiator(ApplicationContext context) {
         return new SpringHandlerInstantiator(context.getAutowireCapableBeanFactory());
     }
+
+    @Override
+    protected void configureMessageConverters(List<HttpMessageConverter<?>> converters) {
```
MiguelMartinRojas: I am not sure why are you using this. Where can I have a look to this when it is activated?

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/RestConfiguration.java
JulianSchallenmueller: What problems are caused by it when you try to run it?

MiguelMartinRojas: I usually run the taskana-spring-example running the main class of the project, I we extends webMvcConfigurationSupport The application doesn´t work when I try to open it in the browser.

PR #614, opened 2018-11-30
TSK-743: Removed HATEOAS from classification- and workbasketdefinition exports

JulianSchallenmueller:


rest/taskana-rest-spring/src/main/java/pro/taskana/rest/ClassificationDefinitionController.java
```diff
@@ -45,19 +43,12 @@
 
     @GetMapping
     @Transactional(readOnly = true, rollbackFor = Exception.class)
-    public ResponseEntity<List<ClassificationResource>> getClassifications(
-        @RequestParam(required = false) String domain) throws ClassificationNotFoundException, NotAuthorizedException,
-        ClassificationAlreadyExistException, ConcurrencyException, DomainNotFoundException, InvalidArgumentException {
+    public ResponseEntity<List<ClassificationSummary>> getClassifications(
```
MiguelMartinRojas: I would rename getClassiciations to exportClassifications like workbasket.

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/ClassificationDefinitionController.java
```diff
@@ -45,19 +43,12 @@
...
-        for (ClassificationSummary summary : summaries) {
-            Classification classification = classificationService.getClassification(summary.getKey(),
-                summary.getDomain());
-            export.add(classificationResourceAssembler.toResource(classification));
-        }
-        return new ResponseEntity<>(export, HttpStatus.OK);
+        return new ResponseEntity<>(summaries, HttpStatus.OK);
     }
 
     @PostMapping(path = "/import")
```
MiguelMartinRojas: - You have to change the object type of import classifications from ClassificationResource to Classfication since we are no longer exporting classificationResources, apply the same to workbasket exportation

- Also in my opinion we are doing a REST API so the path of the Post method should be /import it should be empty.
The same applies to workbasket

PR #615, opened 2018-12-04
TSK-752 Add file uploading component while uploading a classification or workbasket

MiguelMartinRojas:


MiguelMartinRojas:
I have added:

- Block the IU when uploading file 
- Show an error message in case of error.

PR #620, opened 2018-12-18
TSK-756 Add externalId to task to achieve idempotency

BerndBreier:


lib/taskana-core/src/test/java/pro/taskana/impl/TaskServiceImplTest.java
```diff
@@ -1333,9 +1336,44 @@ public void testUpdateTaskRemovingAttachment() throws TaskNotFoundException, Sys
...
 
+    @Test
+    public void testTaskSummaryEqualsHashCode() throws InterruptedException  {
+        Classification classification = createDummyClassification();
+        Workbasket wb = createWorkbasket("WB-ID", "WB-Key");
+        Attachment attachment = JunitHelper.createDefaultAttachment();
+        ObjectReference objectReference = JunitHelper.createDefaultObjRef();
+        TaskImpl taskBefore = createUnitTestTask("ID", "taskName", wb.getKey(), classification);
+        taskBefore.setPrimaryObjRef(objectReference);
+        Thread.sleep(10);
```
MiguelMartinRojas: Why? is there any job working?

lib/taskana-core/src/test/java/pro/taskana/impl/integration/TaskServiceImplIntAutocommitTest.java
```diff
@@ -359,6 +360,7 @@ public void shouldNotTransferByFailingSecurity() throws WorkbasketNotFoundExcept
         task.setId("");
         task.getWorkbasketSummaryImpl().setId(wbNoTransfer.getId());
         task.setWorkbasketKey(null);
+        task.setExternalId(IdGenerator.generateWithPrefix("TST"));
```
MiguelMartinRojas: Why are you using different prefix?

lib/taskana-core/src/test/java/pro/taskana/impl/integration/TaskServiceImplIntExplicitTest.java
```diff
@@ -449,6 +450,7 @@ public void shouldNotTransferByFailingSecurity() throws WorkbasketNotFoundExcept
         task.setId("");
         task.setWorkbasketKey(wbNoTransfer.getKey());
         task.getWorkbasketSummaryImpl().setId(wbNoTransfer.getId());
+        task.setExternalId(IdGenerator.generateWithPrefix("TST"));
```
MiguelMartinRojas: Why are you using different prefix?

rest/taskana-rest-spring-example/pom.xml
```diff
@@ -132,6 +139,11 @@
 			<artifactId>taskana-web</artifactId>
 			<version>${project.version}</version>
 		</dependency>
+		<dependency>
```
MiguelMartinRojas: Leave this dependency for now. But please create a task in taskana board to remove this dependency when taskana-simple-history is released to v.0.0.7

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
```diff
@@ -1124,36 +1160,41 @@ private PrioDurationHolder handleAttachments(TaskImpl task) throws InvalidArgume
...
-                        maxPrio = newPrioDuraton.getPrio();
-                        minDuration = newPrioDuraton.getDuration();
-                    }
-                }
+                actualPrioDuration = handleNonNullAttachment(actualPrioDuration, attachment);
             }
         }
-        if (minDuration != null && MAX_DURATION.equals(minDuration)) {
-            minDuration = null;
+        if (actualPrioDuration.getDuration() != null && MAX_DURATION.equals(actualPrioDuration.getDuration())) {
```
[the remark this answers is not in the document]
BerndBreier: That's correct. I changed it.

lib/taskana-core/src/test/java/pro/taskana/impl/TaskServiceImplTest.java
BerndBreier: In order to guarantee that the 'created' timestamps are different. That's the only data that distinguishes taskBefore and taskAfter ...

lib/taskana-core/src/test/java/pro/taskana/impl/integration/TaskServiceImplIntAutocommitTest.java
BerndBreier: Just to see that this works too. In the final product, only the external ids that were not set by the client (i.e. that are set implicitely by Taskana) will have the prefix "EID:". Currently the external ids supplied by the client don't have a prefix. There may be scenarios where it is beneficial to have a simple means to distinguish internally and externally supplied 'external_id's . In these cases, we could tag external_ids that really refer to an external id with a separate prefix ...

lib/taskana-core/src/test/java/pro/taskana/impl/integration/TaskServiceImplIntExplicitTest.java
BerndBreier: See my last comment

rest/taskana-rest-spring-example/pom.xml
BerndBreier: created TSK-763

lib/taskana-core/src/main/java/pro/taskana/impl/TaskImpl.java
```diff
@@ -849,271 +839,37 @@ public boolean equals(Object obj) {
...
+            workbasketSummary};
+
+            Object[] otherFields = {other.externalId, other.attachments, other.businessProcessId, other.claimed, other.classificationSummary, other.completed,
+            other.created, other.creator, other.custom1, other.custom10, other.custom11, other.custom12, other.custom13, other.custom14,
+            other.custom15, other.custom16, other.custom2, other.custom3, other.custom4, other.custom5, other.custom6, other.custom7,
+            other.custom8, other.custom9, other.customAttributes, other.callbackInfo, other.description, other.due, other.id, other.modified,
+            other.name, other.note, other.owner, other.parentBusinessProcessId, other.planned, other.primaryObjRef, other.state, other.workbasketSummary};
+
+        if (myFields.length != otherFields.length) {
+            throw new SystemException("TaskSummaryImpl: length mismatch between internal arrays");
```
holgerhagen: I don't think that equals() should throw any exception. I would prefer to do a debug statement and return false.

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
```diff
@@ -63,9 +63,19 @@
  */
 public class TaskServiceImpl implements TaskService {
 
+    private static final String IS_ALREADY_CLAIMED_BY = " is already claimed by ";
```
holgerhagen: Just because I'm curious: is this still a good practice or is it "old style" because the Java compiler needed it in earlier versions?

lib/taskana-core/src/test/java/acceptance/task/CreateTaskAccTest.java
```diff
@@ -82,6 +82,51 @@ public void testCreateSimpleManualTask()
...
+
+        try {
+            newTask = taskService.newTask("USER_1_1", "DOMAIN_A");
+            newTask.setExternalId("MyExternalId");
+            newTask.setClassificationKey("T2100");
+            newTask.setPrimaryObjRef(createObjectReference("COMPANY_A", "SYSTEM_A", "INSTANCE_A", "VNR", "1234567"));
+            taskService.createTask(newTask);
+            assertTrue("This line of code must not be reached", false);
+        } catch (TaskAlreadyExistException ex) {
+            assertTrue("correctly enforced Idempotency", true);
```
holgerhagen: better expected=TaskAlreadyExistException ?

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
BerndBreier: It is still a best practice to replace literals that are used in several places by a constant. By the way, this change (as all refactoring changes I made) was requested by SonarQube...

lib/taskana-core/src/test/java/acceptance/task/CreateTaskAccTest.java
BerndBreier: Agreed. See commit TSK-756b

BerndBreier: I also agree that we should have smaller pull requests. But then we should agree that each task is accompanied by a restructuring task that improves the code structure of the files that were touched via the original task.
Because it is a best practice with bug fixing / implementing new features to improve the code that is touched. Ideally, each class touched should be left cleaner than it was before the change...

PR #621, opened 2018-12-19
TSK-747 create query history screen

MiguelMartinRojas:


rest/taskana-rest-spring-example/pom.xml
```diff
@@ -82,6 +46,11 @@
 			<artifactId>taskana-history-rest-spring</artifactId>
 			<version>${project.version}</version>
 		</dependency>
+		<dependency>
```
holgerhagen: I think we agreed, that we want to have the default without history, didn't we?

PR #633, opened 2019-01-22
TSK-770 Refactor hateoas generation adding spring JAspect support

MiguelMartinRojas:
- Refactored HATEOAS resource generation
- Unify Resources naming/creation/structure...
- Unify Assemblers creation extending ResourceAssemberSupport
- Add JAspect for pagination links and information generation
- Remove assemblers for list.
- Created PageLink annotation
- Avoid BeanUtils.copyProperties usage since it is using reflection and is low performance
- Encapsulate resources implementation into their classes instead of assemblers 
- Use Spring dependency injection properly for assemblers in controllers.


mustaphazorgati:
You probably squashed by accident the travis commit where it updates the poms with your commit. Please remote those changes to the version and rename your commit message.

PR #636, opened 2019-01-28
Task/685/692 Added DailyEntryExitReport grouped by org level. Backend only (1/2)

mustaphazorgati:


lib/taskana-core/src/main/java/pro/taskana/impl/DailyEntryExitReportBuilderImpl.java
```diff
@@ -0,0 +1,95 @@
...
+    @Override
+    protected DailyEntryExitReport.Builder _this() {
+        return this;
+    }
+
+    // since this method is not documented I have no idea what the proper groupedBy should be,
+    // thus this method is not supported on this builder.
+    @Override
+    public List<String> listTaskIdsForSelectedItems(List<SelectedItem> selectedItems) {
+        throw new UnsupportedOperationException("");
```
mustaphazorgati: Find out what this does.

lib/taskana-core/src/main/java/pro/taskana/impl/TimeIntervalReportBuilderImpl.java
```diff
@@ -64,25 +68,25 @@ public B inWorkingDays() {
 
     @Override
     public B workbasketIdIn(List<String> workbasketIds) {
-        this.workbasketIds = workbasketIds;
+        this.workbasketIds = new ArrayList<>(workbasketIds);
```
mustaphazorgati: have a look into Collections.unmodifiableList()

lib/taskana-core/src/test/java/pro/taskana/impl/ClassificationServiceImplTest.java
```diff
@@ -45,24 +44,22 @@
 @RunWith(MockitoJUnitRunner.class)
 public class ClassificationServiceImplTest {
 
-    private final String todaysDate = LocalDate.now().toString().substring(0, 10);
+    // This is important. Since ClassificationService.createClassification() uses an Instant for the time
+    // and thus the UTC time zone. e.g. LocalDate uses the system timezone. This may lead to issues when the day differs
+    // between those timezones.
+    private final String todaysDate = Instant.now().toString().substring(0, 10);
```
mustaphazorgati: Don't use string to compare Dates.

lib/taskana-core/src/main/java/pro/taskana/impl/TimeIntervalReportBuilderImpl.java
mustaphazorgati: Collections.unmodifiableList() makes the current reference unmodifiable, but the given list can still be modified.

PR #637, opened 2019-01-28
TSK-774: set Flags for listValues()

BVier:
Removed callback_info and custom_attributes because DB2 can't ORDER BY an CLOB value and they aren't necessary.


lib/taskana-core/src/main/java/pro/taskana/mappings/QueryMapper.java
```diff
@@ -672,8 +672,8 @@
         + "group by WORKBASKET_ID ) AS f where max_read = 1 ) "
         + "</if> "
         + "<if test='taskIds != null'>AND t.ID IN(<foreach item='item' collection='taskIds' separator=',' >#{item}</foreach>)</if> "
-        + "<if test='externalIdIn != null'>AND a.EXTERNAL_ID IN(<foreach item='item' collection='externalIdIn' separator=',' >#{item}</foreach>)</if> "
-        + "<if test='externalIdLike != null'>AND (<foreach item='item' collection='externalIdLike' separator=' OR '>UPPER(a.EXTERNAL_ID) LIKE #{item}</foreach>)</if> "
+        + "<if test='externalIdIn != null'>AND t.EXTERNAL_ID IN(<foreach item='item' collection='externalIdIn' separator=',' >#{item}</foreach>)</if> "
```
BVier: Changed this because attachment doesn't have such an attribute. So it's probably a typo.
In my opinion it wouldn't make sense to do an own pull request for this.

lib/taskana-core/src/main/java/pro/taskana/impl/TaskQueryImpl.java
```diff
@@ -1014,14 +1014,16 @@ private void setupAccessIds() {
...
             checkOpenAndReadPermissionForSpecifiedWorkbaskets();
             setupAccessIds();
+
             if (columnName.equals(TaskQueryColumnName.CLASSIFICATION_NAME)) {
                 joinWithClassifications = true;
-                addClassificationNameToSelectClauseForOrdering = true;
             }
-            if (columnName.equals(TaskQueryColumnName.A_CLASSIFICATION_NAME)) {
-                joinWithAttachments = true;
+            List<TaskQueryColumnName> attachmentColumns = Arrays.asList(
```
TitusLabs: I would prefer it to select them by convention like all columns starting with A_
You will not adapt this while adding a new column to the enum.
furthermore this is a static value which could be extracted. I would actually include it in the enum class itself.

BVier: Good point!
Also didn't liked my solution at this point, but I wasn't aware that I could add such a function to a Enum class.
Thanks for reporting!

BerndBreier:
This pull request is replaced by PR "B vier tsk 772 lazy #641" which fixes problems detected during travis build / test of this PR.

PR #639, opened 2019-01-30
TSK-778 - Check if exist the history_event's table before running the script

ignaciorecuerda:


rest/taskana-rest-spring-example/src/main/resources/sql/sample-data/check-history-event-exist.sql
```diff
@@ -0,0 +1,3 @@
+SET SCHEMA TASKANA;
```
MiguelMartinRojas: TASKANA should be variable.

rest/taskana-rest-spring-example/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
```diff
@@ -110,12 +113,26 @@ public void generateSampleData(String schemaName) {
         }
 
         runner.setStopOnError(true);
+
+        String[] script = {WORKBASKET, DISTRIBUTION_TARGETS, CLASSIFICATION, TASK, ATTACHMENT, WORKBASKET_ACCESS_LIST,
```
MiguelMartinRojas: Move this logic into a function like getScriptsList

rest/taskana-rest-spring-example/src/main/resources/sql/sample-data/check-history-event-exist.sql
ignaciorecuerda: In this case, this property is unnecessary.
Removed!

MiguelMartinRojas: no, it should be variable, not removed. find schemaName into the taskana-core

PR #648, opened 2019-02-18
TSK-799 - Refactor REST documentation

ignaciorecuerda:
- Create a new module which contains all documentation test and then, the documentation is copied to taskana-rest-spring-example and taskana-rest-spring-wildfly-example-
- Fix the history documentation (the links was wrong)


MiguelMartinRojas:
- Remove applicaiton.properties from taskana-history-rest-spring/test/resources.
- Do you need taskana-history-rest-spring/main/resources/ folder and it's content?
- Review dependecies from taskana-history-rest-spring/pom.xml
- taskana-rest-documentation doesn't look like an extandar spring boot module 
- Review taskana-rest-documentation resources of test and java/rest are all of them needed?
- Can be shared copy-documentation-to-taskana-rest-spring-example in taskana-rest-documentation pom.xml? you are duplicating this execution
- Move favicon.ico from taskana-rest-spring-example-resources/static/img to rest-spring-setup

ignaciorecuerda:
- The module taskana-history-rest-spring was removed
- Removed the unnecessary resources in both test and java part
- we need two execution in taskana-rest-documentation pom.xml file because in the same execution, it can't copy the resource to two different locations
- favicon.ico was removed. Now it used logo.png which is also used in the frontend part

PR #659, opened 2019-03-07
task/819 Split travis into 6 jobs with a more understandable name

MiguelMartinRojas:


ci/test.sh
```diff
@@ -26,13 +31,17 @@ function main {
   [[ $# -eq 0 || "$1" == '-h' || "$1" == '--help' ]] && helpAndExit 0
```
mustaphazorgati: please check if there exist 2 parameters.

ci/test.sh
```diff
@@ -15,6 +15,11 @@ set -e # fail fast
 #H   - DB2_10_5
 #H   - DB2_11_1
 #H   - POSTGRES_10_4
+#H project:
```
mustaphazorgati: please update in usage aswell (line 9)

ci/test.sh
```diff
@@ -26,13 +31,17 @@ function main {
   [[ $# -eq 0 || "$1" == '-h' || "$1" == '--help' ]] && helpAndExit 0
   REL=`dirname "$0"`
   eval "$REL/prepare_db.sh '$1'"
-  if [[ "$1" == "H2" ]]; then
-    (cd $REL/../web && npm run test)
-    mvn clean verify -q -f $REL/../lib/ -B
-    mvn clean install -q -f $REL/../rest/ -B -P history.plugin
-  elif [[ "$1" == "POSTGRES_10_4" ]]; then
+  if [[ "$1" == "H2" && "$2" == "LIB" ]]; then
+    mvn clean install -q -f $REL/../lib/ -B
```
mustaphazorgati: is there a reason why we install the artifacts again?

ci/test.sh
```diff
@@ -26,13 +31,17 @@ function main {
...
+  elif [[ "$1" == "H2" && "$2" == "REST" ]]; then
+   (cd $REL/../web && npm run test)
+    mvn clean install -q -f $REL/../rest/ -B
```
mustaphazorgati: please add a comment explaining we need to package rest again because our rest documentation is build when running tests

mustaphazorgati:
While you already at it: -DskipTests=true can be reduced to -DskipTests. Same applies for -Dmaven.javadoc.skip

ci/test.sh
MiguelMartinRojas: We are not checking the number of parameter in any of the other files. Do you still want to check the number of parameters for all the scripts files?

ci/test.sh
MiguelMartinRojas: We have to do a complete installation with tests in this stage, the main stage is H2 && REST.
We can skip javadoc in this installation too(I will add it).

ci/test.sh
mustaphazorgati: Just checked your point. True that. Not necessary

PR #665, opened 2019-03-11
task/828: Unify and reduce the response and logger code in the controllers

ignaciorecuerda:


rest/taskana-rest-spring/src/main/java/pro/taskana/rest/ClassificationDefinitionController.java
```diff
@@ -78,11 +78,12 @@
             export.add(classificationResourceAssembler.toDefinition(classification));
         }
```
MiguelMartinRojas: Can you review this file, there are some typos mapChildsToParentKeys Childs doesn´t exists in english.
Also you can remove JsonParseException, JsonMappingException  exceptions from extractClassificationResourcesFromFile method, IOException is shallowing them.

ignaciorecuerda: I reviewed the class and changing the name of some properties. Also, I removed JsonParseException and JsonMappingException

PR #676, opened 2019-03-19
TSK-844: Reorder columns in history view and sort by "created"-value

BVier:
Reorders the columns and add the "created"-Instant (works as timestamp).
Default ordering is by "created" ascending.

A translation for the date-query is still missing.
It should be translated into a "TimeInterval"-Entity of the History-Service

### DIFF
--- a/web/src/app/history/task-query/task-query.component.html
+++ b/web/src/app/history/task-query/task-query.component.html
@@ -39,8 +39,11 @@
           </div>
           <div class="divTableRow">
             <ng-container *ngFor="let taskHeader of taskQueryHeader | mapToIterable">
-              <div class="divTableHeader" *ngIf="filterFieldsToShow(taskHeader.key) && !filterExpandGroup(taskHeader.key)">
-                <input type="text" (keyup.enter)="search()" class="form-control input-sm" placeholder="{{getHeaderFieldDescription(taskHeader.key)}}"
+              <div class="divTableHeader divDate" *ngIf="isDate(taskHeader.key)">
+                <taskana-date-picker type="text" placeholder="{{getHeaderFieldDescription(taskHeader.key)}}" [name]="'this.created'" [id]="'created'" (dateOutput)="updateDate($event)" formControlName="{{taskHeader.key}}"></taskana-date-picker>
+              </div>
+              <div class="divTableHeader" *ngIf="!isDate(taskHeader.key) && filterFieldsToShow(taskHeader.key) && !filterExpandGroup(taskHeader.key)">
+                <input type="text" class="form-control input-sm" (keyup.enter)="search()" placeholder="{{getHeaderFieldDescription(taskHeader.key)}}"
                   formControlName="{{taskHeader.key}}">
               </div>
               <div class="divTableHeader divTableHeaderSeparator" [ngClass]="{'zoom-in': !toggleExpand, 'zoom-out': toggleExpand}"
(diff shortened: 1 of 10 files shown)


web/src/app/history/task-query/task-query.component.html
```diff
@@ -56,7 +56,10 @@ <h4 class="panel-header">Taskana history query</h4>
         <div class="divTableBody">
           <div *ngFor="let task of taskQuery" class="divTableRow">
             <ng-container *ngFor="let taskHeader of taskQueryHeader | mapToIterable">
-              <div class="divTableCell" *ngIf="filterFieldsToShow(taskHeader.key) && !filterExpandGroup(taskHeader.key)">
+              <div class="divTableCell" *ngIf="isDate(taskHeader.key)">
+                {{task[taskHeader.key] | date:'yyyy-MM-dd'}}
```
MiguelMartinRojas: I would use date-time-zone.pipe.ts instead of date pipe. But date-time-zone.pipe.ts is defined in task/831 so you will have to wait until it is merged.

MiguelMartinRojas: Make  sure that the task number matches with jira task number here you have TSK-838 but in jira TSK-842

BVier: I can change it, but I don't think I need the 'HH:mm:ss'-part.
For me it's easier to understand when I don't see the time.

MiguelMartinRojas: Benni, it's not just showing the correct date format, date-time-zone pipe applies timezone according to the browser location, if you want yo can extend the pipe for adding the date format. but this is up to you.

BVier: Ok, now I understood.
I try to change it!

web/src/app/shared/pipes/date-time-zone/date-time-zone.pipe.ts
```diff
@@ -8,11 +8,15 @@ export class DateTimeZonePipe implements PipeTransform {
...
-  transform(value: any, args?: any): any {
+  transform(value: any, displayTime: boolean = true, args?: any): any {
 
     let date = this.datesMap.get(value);
     if (!date) {
-      date = TaskanaDate.getDateToDisplay(value);
+      if (displayTime) {
+        date = TaskanaDate.getDateToDisplay(value);
+      } else {
+        date = TaskanaDate.convertSimpleDate(new Date (value));
```
MiguelMartinRojas: Convert simple date doesn't apply timeZone, if you have a look to getDateToDisplay it applies time zone explicitly.
you have to change getDateToDisplay method adding a new argument, dateFormat for instance and apply the format desired to applyTimeZone,

this is a suggestion:

`
public static dateFormatWithTime: string = 'yyyy-MM-dd HH:mm:ss'
public static dateFormat: string = 'yyyy-MM-dd'

public static getDateToDisplay(date: string, dateFormat: string = this.dateFormatWithTime): string {
    return this.applyTimeZone(date, dateFormat);
  }

  public static applyTimeZone(date: string): string | null {
    const dateFormat = 'yyyy-MM-dd HH:mm:ss';
    const dateLocale = 'en-US';
    const datePipe = new DatePipe(dateLocale);
    return datePipe.transform(date, dateFormat, Intl.DateTimeFormat().resolvedOptions().timeZone);
  }
`

web/src/app/shared/util/taskana.date.ts
```diff
@@ -13,7 +13,7 @@ export class TaskanaDate {
     const dateFormat = 'yyyy-MM-dd';
     const dateLocale = 'en-US';
     const datePipe = new DatePipe(dateLocale);
-    return datePipe.transform(date, dateFormat);
+    return datePipe.transform(date, dateFormat, Intl.DateTimeFormat().resolvedOptions().timeZone);
```
MiguelMartinRojas: This is no needed. const datePipe = new DatePipe(dateLocale); already applies timezone

BVier:
Everything works, but the created field is a bit larger due to settings for the date-picker

web/src/app/shared/util/taskana.date.ts
MiguelMartinRojas: Rollback this line

PR #681, opened 2019-03-28
bug/837 Get implementation version from package meta information instead of spring configuration

MiguelMartinRojas:


lib/taskana-core/pom.xml
```diff
@@ -1,303 +1,314 @@
 <project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
```
MiguelMartinRojas: I only have added this plugin, the rest is autoformatting.
<plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-jar-plugin</artifactId>
                <configuration>
                    <archive>
                        <manifest>
                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
                        </manifest>
                    </archive>
                </configuration>
            </plugin>

PR #684, opened 2019-04-03
task/763 remove web dependency from taskana-rest-spring-example

MiguelMartinRojas:


MiguelMartinRojas:
It was already merged ->  https://github.com/Taskana/taskana/pull/678/files

PR #692, opened 2019-05-06
TSK-847: Add Authentication to Import-Export-Module

BVier:


BVier:
Hi Mustapha,
of course I used the devMode!
Export works fine but import not.

I see the bigger problem in the export module (where no authentication is required) than with the import module.

PR #697, opened 2019-05-09
task/853 [UI] Erroneous navigation when changing domains

ignaciorecuerda:


ignaciorecuerda:
The explanation is that now the router in Taskana start in /taskana, so if it doesn't find the route the application will redirect to the default route. You can find this in https://github.com/Taskana/taskana/blob/master/web/src/app/app-routing.module.ts
For example, now the route for the workplace is "taskana/workplace" before it was "workplace"

PR #698, opened 2019-05-10
TSK-850 restructured poms + TSK-845 removed compiler warnings

mustaphazorgati:
What was done:

- created taskana-parent 
- moved all properties to taskana-parent
- moved all sonatype requirements to taskana-parent (except for name and description)
- enabled checkstyle for every module
- fixed checkstyle errors
- removed all build errors
- enabled failOnWarning for future builds
- test deployment was done before rebasing (https://travis-ci.org/mustaphazorgati/taskana/builds/530679827)

@holgerhagen, @TitusLabs While fixing checkstyle I made some JavaDoc comments which only contain the word "TODO". I did this because I was not sure what to write. Please either comment what you would write as a comment or create a new commit where you replace "TODO" with the actual comment. - Thanks!


TitusLabs:
this pr is mixed up with other changes. but good job. there is too much copy&paste in the poms. but, i would go for a bom instead of using version properties in every pom. just keep the properties in the bom and you can omit the version tags in all other poms..

mustaphazorgati:
> good job! there is too much copy&paste in the poms. but, i would go for a bom instead of using version properties in every pom. just keep the properties in the bom and you can omit the version tags in all other poms..

What I can do is find out which dependencies are used more than 2 times and put those inside `dependencyManagement` in the taskana-parent pom

TitusLabs:
> What I can do is find out which dependencies are used more than 2 times and put those inside `dependencyManagement` in the taskana-parent pom

for example.. but i would do this with all used dependencies. so you have them and their versions on one place.

mustaphazorgati:
I can do that, but that would result in copying the dependencies into the parent pom. Removing a dependency will then be more complicated since you have to check if a second pom is importing the same dependency. Using the properties we would not have the issue (+ our IDE would tell us if a property is not used)

mustaphazorgati:
But if I can find boms for a lot of dependencies it would change.. I'll have a look

rest/taskana-rest-spring-example/src/main/resources/application.properties
```diff
@@ -52,3 +52,5 @@ spring.resources.cache.cachecontrol.cache-private=true
 ####### tomcat is not detecting the x-forward headers from bluemix as a trustworthy proxy
 server.tomcat.internal-proxies=.*
 server.use-forward-headers=true
+####### allow bean overriding
+spring.main.allow-bean-definition-overriding=true
```
holgerhagen: What do we need this for?

lib/pom.xml
```diff
@@ -1,22 +1,25 @@
...
-        <modelVersion>4.0.0</modelVersion>
-        <groupId>pro.taskana</groupId>
-        <artifactId>taskana-parent</artifactId>
-        <version>1.1.3-SNAPSHOT</version>
-        <packaging>pom</packaging>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<artifactId>taskana-lib-parent</artifactId>
+	<version>1.1.3-SNAPSHOT</version>
```
holgerhagen: Eclipse recommends to have the verion in the parent only. This is just duplicate information. And I agree.

lib/taskana-cdi/pom.xml
```diff
@@ -179,22 +53,25 @@
...
 			<scope>test</scope>
 		</dependency>
 		<!-- Brought in via WildFly Swarm `bom` -->
 		<dependency>
 			<groupId>org.wildfly.swarm</groupId>
 			<artifactId>arquillian</artifactId>
+			<version>${version.wildfly.swarm}</version>
 			<scope>test</scope>
 		</dependency>
 		<!-- Brought in via Arquillian BOM, see dependencyManagement section above -->
```
holgerhagen: ... but there is no dependencyManagement section anymore, is there?

lib/pom.xml
```diff
@@ -1,22 +1,25 @@
...
-        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-        <modelVersion>4.0.0</modelVersion>
-        <groupId>pro.taskana</groupId>
-        <artifactId>taskana-parent</artifactId>
-        <version>1.1.3-SNAPSHOT</version>
-        <packaging>pom</packaging>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<artifactId>taskana-lib-parent</artifactId>
```
holgerhagen: The Eclipse project is still named "taskana-parent", which causes some irritations now. Is it possible to change the name of the Eclipse project as well? Would it make sense to change the folders as well to be consistent?

rest/taskana-rest-spring-example/src/main/resources/application.properties
mustaphazorgati: This is enabled in the `spring-boot-starter-parent`. Since I replaced the parent with our `taskana-rest-parent` I have to manually enable overriding. Having it disabled throws an error while starting the example app..

I just tried it out, somehow this is not necessary anymore..

lib/pom.xml
mustaphazorgati: Yes and no.
It is duplicate information, but since the release modifies the version the duplicate information is necessary. Otherwise it throws an error because the version is inherited from the parent.
See: https://travis-ci.org/mustaphazorgati/taskana/jobs/530279994

lib/taskana-cdi/pom.xml
mustaphazorgati: true, forgot to remove it.

lib/pom.xml
mustaphazorgati: This seems to be a local configuration error. I removed all modules and reimportet them. Seems to be fine
Screenshot from 2019-05-17 18-43-27

lib/pom.xml
holgerhagen: My understanding would be that we have to change the version on the parent, only, and the child modules are updated automatically. This is not true?

See https://stackoverflow.com/questions/5726291/updating-version-numbers-of-modules-in-a-multi-module-maven-project

lib/pom.xml
holgerhagen: You are right. This is stored in .project which is not checkin. I will have to correct this manually.

rest/taskana-rest-spring-example/src/main/resources/application.properties
holgerhagen: So you are going to remove it? Or do I get you wrong here?

If we still need it: what is the exception?

lib/pom.xml
mustaphazorgati: That is true. But since we are deploying the child modules by themself the version tag is necessesary because the mvn deploy command tries to change the version.

rest/taskana-rest-spring-example/src/main/resources/application.properties
mustaphazorgati: yes. I will delete it.

PR #702, opened 2019-05-27
TSK-671: Allow upload of large classification or workbasket files

BVier:
Allow larger uploads (up to 10 MB).
Additionally corrected a typo.

### DIFF
--- a/rest/taskana-rest-spring-example/src/main/resources/application.properties
+++ b/rest/taskana-rest-spring-example/src/main/resources/application.properties
@@ -49,6 +49,9 @@ taskana.ldap.groupsOfUser=memberUid
 taskana.jobscheduler.async.cron=0 * * * * *
 ####### cache static resources properties
 spring.resources.cache.cachecontrol.cache-private=true
+####### for upload of big workbasket- or classification-files
+spring.servlet.multipart.max-file-size=10MB
+spring.servlet.multipart.max-request-size=10MB
 ####### tomcat is not detecting the x-forward headers from bluemix as a trustworthy proxy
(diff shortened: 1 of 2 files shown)


rest/taskana-rest-spring-example/src/main/resources/application.properties
```diff
@@ -49,6 +49,9 @@ taskana.ldap.groupsOfUser=memberUid
 taskana.jobscheduler.async.cron=0 * * * * *
 ####### cache static resources properties
 spring.resources.cache.cachecontrol.cache-private=true
+####### for upload of big workbasket- or classification-files
+spring.servlet.multipart.max-file-size=10MB
```
mustaphazorgati: is 10mb a reasonable size? I thing 100mb or something around there is more reasonable.

BVier: The whole classification file for the db test-data is about 1,5MB.
So I thought 10 MB are more than enough space for at least this case.

The workbasket file in our db is even smaller, but probably it's bigger in the real use case.
Still it should not be bigger than 10MB in my opinion.

But of course we don't know how the file size will change for future use cases.
In this case the admin could simply change this file.

BVier: But if there are no security concerns we could set the file-size to '-1' and we don't have to decide, what the 'real' maximum size is.

holgerhagen: It's a good practice to restrict the size. Let's stick to the 10 MB and see how far we can get with it.

PR #703, opened 2019-05-27
TSK-859: make unlinking child from parent through import possible

BVier:
When uploading an existing child-classification where the parent was removed, the parent-properties were not updated and the updated classification was still a child.

### DIFF
--- a/rest/taskana-rest-spring-test/src/test/java/pro/taskana/rest/ClassificationDefinitionControllerIntTest.java
+++ b/rest/taskana-rest-spring-test/src/test/java/pro/taskana/rest/ClassificationDefinitionControllerIntTest.java
@@ -337,6 +337,37 @@ public class ClassificationDefinitionControllerIntTest {
         assertNotEquals(wrongParentCl.getClassificationId(), childCl.getParentId());
     }
 
+    @Test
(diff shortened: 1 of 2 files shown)


rest/taskana-rest-spring-test/src/test/java/pro/taskana/rest/ClassificationDefinitionControllerIntTest.java
```diff
@@ -337,6 +337,38 @@ public void testImportParentAndChildClassificationWithKey() throws IOException {
...
+        child2.setParentKey("");
+        String withoutParent = objMapper.writeValueAsString(child2);
+
+        List<String> clList = new ArrayList<>();
+        clList.add(withNewParent);
+        clList.add(withoutParent);
+
+        ResponseEntity<String> response = importRequest(clList);
+        assertEquals(HttpStatus.OK, response.getStatusCode());
+        Thread.sleep(10000);
```
mustaphazorgati: Why 10 seconds? I think this is unnecessary. Tested it locally with 10ms. Runs fine.

BVier: You're right!
Thanks for your testing!

PR #704, opened 2019-05-29
TSK-835: enabled hashrouting to find resources from a context path

mustaphazorgati:


TitusLabs:
ah, nice.. as I expected it :-)

PR #707, opened 2019-07-18
Task/865 Separated TaskanaEngine into an internal and external engine

mustaphazorgati:
For easier understanding:
[taskana-poc.pdf](https://github.com/Taskana/taskana/files/3406058/taskana-poc.pdf)

### DIFF
--- a/lib/taskana-core/src/main/java/pro/taskana/TaskanaEngine.java
+++ b/lib/taskana-core/src/main/java/pro/taskana/TaskanaEngine.java
@@ -2,11 +2,8 @@ package pro.taskana;
 
 import java.sql.SQLException;
 
-import org.apache.ibatis.session.SqlSession;
-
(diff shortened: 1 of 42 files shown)


mustaphazorgati:
PR is currently on hold.

lib/taskana-core/src/main/java/pro/taskana/TaskanaEngine.java
```diff
@@ -52,6 +55,12 @@
      */
     TaskanaEngineConfiguration getConfiguration();
 
+    /**
+     * The HistoryEventProducer events and emits them to the registered history service providers.
+     * @return The HistoryEventProducer
+     */
+    HistoryEventProducer getHistoryEventProducer();
```
mustaphazorgati: This is supposed to be an internal operation. Please move it to the Internal interface

PR #714, opened 2019-09-02
[CLEANUP] moved  eclipse specific setup into seperated profile to reduce build warnings

mustaphazorgati:
…uce build warnings


mustaphazorgati:
BEFORE MERGING:
Inform all contributer to use the maven profile `eclipse` when they develop with Eclipse IDE and update our doc.

PR #715, opened 2019-09-10
TSK-873: Support for Maven 3.6.x

lehmanndennis:


ci/test.sh
```diff
@@ -33,24 +33,24 @@ function main {
...
+    (cd $REL/../lib/ && mvn clean install -q -B -DskipTests -Dmaven.javadoc.skip)
     
-    mvn clean install -q -f $REL/../rest/ -B -DskipTests -pl !taskana-rest-spring-wildfly-example -Dmaven.javadoc.skip
-    mvn clean install -q -f $REL/../rest/ -B -pl taskana-rest-spring-wildfly-example -Dmaven.javadoc.skip -P postgres
+    (cd $REL/../rest/ && mvn clean install -q -B -DskipTests -pl !taskana-rest-spring-wildfly-example -Dmaven.javadoc.skip)
+    (cd $REL/../rest/ && mvn clean install -q -B -pl taskana-rest-spring-wildfly-example -Dmaven.javadoc.skip -P postgres)
   else
-    mvn clean verify -q -f $REL/../lib/taskana-core -B
+    (cd $REL/../lib/taskana-core && mvn clean verify -q -B)
   fi
```
mustaphazorgati: just out of curiosity: Why you change this? 
According to the help page of maven the -f command takes the pom file from a given directory

lib/taskana-cdi/src/test/java/pro/taskana/TaskanaProducersTest.java
```diff
@@ -33,19 +31,13 @@
         deployment.addAllDependencies();
         deployment.addDependency("org.mybatis:mybatis:3.4.2");
         deployment.addDependency("org.mybatis:mybatis-cdi:1.0.0");
-        deployment.addDependency("pro.taskana:taskana-core:1.1.4-SNAPSHOT");
+        deployment.addDependency("pro.taskana:taskana-core");
```
mustaphazorgati: What does this change exactly do? Will the latest version of taskana-core be used?
If so, then please remove the -swarm parameter from the change_version call in .travis.yml

ci/test.sh
lehmanndennis: I changed it because there is a bug in Wildfly Swarm / Thorntail (https://issues.jboss.org/browse/THORN-2049) and with the -f flag the tests would fail. For convenience I changed the -f flag everywhere to the cd command

lib/taskana-cdi/src/test/java/pro/taskana/TaskanaProducersTest.java
lehmanndennis: It will use the Version specified in the pom. So the latest version will be used.

.travis.yml
```diff
@@ -36,7 +36,6 @@ script:
   - "sh -e /etc/init.d/xvfb start"
   - sleep 3 # give xvfb some time to start
   - nvm install $NODE_VERSION
-    && ci/change_version.sh -m . -swarm lib/taskana-cdi/src/test/java/pro/taskana/TaskanaProducersTest.java
```
mustaphazorgati: Only the swarm parameter please. We still need to change the version when we have a release build

mustaphazorgati: We also have 3 occurances of the  -swarm parameter. You just removed one of them. :)

ci/test.sh
mustaphazorgati: Alright. Personally I do not like using subshells for building the application. Can you please put this change in a seperate commit so that we can revert it whenever the bug is fixed?

PR #720, opened 2019-09-13
TSK-879: Removed initial mvn build

lehmanndennis:


.travis.yml
```diff
@@ -30,6 +30,8 @@ env:
     - DB=POSTGRES_10_4 PROJECT=CORE
     - DB=POSTGRES_10_4 PROJECT=WILDFLY
 
+install: mvn dependency:resolve
```
mustaphazorgati: Why not skip?

lehmanndennis: Cleaner output from the build scripts. Dependencies are resolved and downloaded already when building the project.

PR #753, opened 2019-11-07
TSK-947: Extract Url mappings

benjamineckstein:


benjamineckstein:
no one cares :( working on a better solution. will be integrated within another PR.

PR #754, opened 2019-11-12
TSK-951: Ignore maven javadocs for web build

benjamineckstein:
image

### DIFF
--- a/web/pom.xml
+++ b/web/pom.xml
@@ -12,7 +12,9 @@
 		<version>1.1.5-SNAPSHOT</version>
 		<relativePath>../rest/pom.xml</relativePath>
 	</parent>
-
+	<properties>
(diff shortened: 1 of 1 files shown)


web/pom.xml
```diff
@@ -39,7 +41,13 @@
 					</execution>
 				</executions>
 			</plugin>
-		</plugins>
+      <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <skipIfEmpty>true</skipIfEmpty>
```
mustaphazorgati: Why do we need this? Jar should never be empty because we have copied the web api in /static

benjamineckstein: [WARNING] JAR will be empty - no content was marked for inclusion!

this will fix this. content moving to static belongs to the validate phase. 
<phase>validate</phase>

PR #756, opened 2019-11-13
TSK-864: Expected exception handling for unit tests

benjamineckstein:


lib/taskana-core/src/test/java/acceptance/task/TransferTaskAccTest.java
```diff
@@ -309,21 +309,15 @@ public void testTransferTasksWithListNotSupportingRemove() throws NotAuthorizedE
...
-        try {
-            taskService.transferTasks("WBI:100000000000000000000000000000000006", null);
-            Assert.fail("exception was excepted to be thrown");
-        } catch (InvalidArgumentException e) {
-            Assert.assertEquals(e.getMessage(), "TaskIds must not be null.");
-        }
+
+        Throwable t = Assertions.assertThrows(InvalidArgumentException.class, () ->
+            taskService.transferTasks("WBI:100000000000000000000000000000000006", null));
+        Assert.assertEquals(t.getMessage(), "TaskIds must not be null.");
```
mustaphazorgati: This is more a question than a comment:
Is it smart to test exception messages? I think this could be a pain if we want to refactor those messages. Maybe a cleanup ticket for the future?

lib/taskana-core/src/test/java/acceptance/workbasket/CreateWorkbasketAccTest.java
```diff
@@ -110,59 +109,45 @@ public void testCreateWorkbasketWithMissingRequiredField()
...
-        workbasket.setOrgLevel1("company");
-        try {  // empty name
-            workbasketService.createWorkbasket(workbasket);
-            fail("InvalidWorkbasketException was expected");
-        } catch (InvalidWorkbasketException e) {
-        }
+        // missing key
+        Assertions.assertThrows(InvalidWorkbasketException.class, () -> workbasketService.createWorkbasket(workbasket));
+
+        Workbasket workbasket2 = workbasketService.newWorkbasket("key", "novatec");
```
mustaphazorgati: why are you not reuising the variable workbasket? I think the number in the name is not aesthetically pleasing.

lib/taskana-core/src/test/java/acceptance/task/DeleteTaskAccTest.java
```diff
@@ -2,7 +2,6 @@
 
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
```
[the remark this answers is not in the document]
benjamineckstein: hmm yes. but that's an easier task once everything is migrated to JUnit 5

lib/taskana-core/src/test/java/acceptance/task/TransferTaskAccTest.java
benjamineckstein: good point. we should use a reason enum within the exception and not rely on some strings ... but that's not part of this ticket ;D

lib/taskana-core/src/test/java/acceptance/workbasket/CreateWorkbasketAccTest.java
benjamineckstein: The test itself is not pleasing and covers way too much ... but the technical reason is that you can only use final or effectively final variables within lambda expressions.

PR #759, opened 2019-11-19
TSK-960: Refactor initTaskanaRoles

benjamineckstein:


lib/taskana-core/src/main/java/pro/taskana/TaskanaRole.java
```diff
@@ -16,17 +22,17 @@
...
-        } else if (MONITOR.propertyName.equalsIgnoreCase(name)) {
-            return TaskanaRole.MONITOR;
-        } else {
-            return null;
-        }
+        return Arrays.stream(TaskanaRole.values())
+            .filter(x -> x.propertyName.equalsIgnoreCase(name))
+            .findFirst()
+            .orElseThrow(() -> new SystemException(
+                "Internal System error when processing role property " + name));
```
mustaphazorgati: Why are you throwing an exception here? This is a new behaviour.

lib/taskana-core/src/main/java/pro/taskana/configuration/TaskanaEngineConfiguration.java
```diff
@@ -298,36 +299,27 @@ private void initSchemaName(String schemaName) {
...
-                    throw new SystemException(
-                        "Internal System error when processing role property " + propertyName);
-                }
-            }
-        }
+        List<String> validPropertyNames = TaskanaRole.getValidPropertyNames();
+
+        props.keySet()
+            .stream()
+            .map(key -> (String) key)
```
mustaphazorgati: I do not like casts. String::valueof

lib/taskana-core/src/main/java/pro/taskana/configuration/TaskanaEngineConfiguration.java
```diff
@@ -298,36 +299,27 @@ private void initSchemaName(String schemaName) {
...
 
         if (LOGGER.isDebugEnabled()) {
             roleMap.forEach(
                 (k, v) -> LOGGER.debug("Found Taskana RoleConfig {} : {} ", k, LoggerUtils.setToString(v)));
         }
+    }
 
+    private HashSet<String> getTokensWithCollection(String str, String rolesSeparator) {
+        return Collections.list(new StringTokenizer(str, rolesSeparator)).stream()
+            .map(token -> ((String) token).toLowerCase().trim())
```
mustaphazorgati: Same as above. I do not like casts. String::valueof

lib/taskana-core/src/main/java/pro/taskana/configuration/TaskanaEngineConfiguration.java
benjamineckstein: great. i like that improvement!

lib/taskana-core/src/main/java/pro/taskana/configuration/TaskanaEngineConfiguration.java
benjamineckstein: ;)

lib/taskana-core/src/main/java/pro/taskana/TaskanaRole.java
benjamineckstein: It was just moved from initTaskanaRoles to prevent returning null. since this is the only method calling that, it does not change anything. 
suggestions? alternatively i could return an optional

mustaphazorgati: I saw what you described later during the review. The solution is fine here. Personally I would have thrown an InvalidArgumentException.

lib/taskana-core/src/main/java/pro/taskana/configuration/TaskanaEngineConfiguration.java
mustaphazorgati: Still the cast :D

lib/taskana-core/src/main/java/pro/taskana/configuration/TaskanaEngineConfiguration.java
mustaphazorgati: Cast is still there :D

PR #760, opened 2019-11-19
TSK-950: Test pojos with openpojo

benjamineckstein:
image


lib/taskana-core/src/main/java/pro/taskana/impl/WorkbasketAccessItemImpl.java
```diff
@@ -548,6 +550,20 @@ public boolean equals(Object obj) {
...
+        } else if (!workbasketKey.equals(other.workbasketKey)) {
+            return false;
+        }
+        if (accessName == null) {
+            if (other.accessName != null) {
+                return false;
+            }
+        } else if (!accessName.equals(other.accessName)) {
+            return false;
+        }
```
benjamineckstein: found 3 fields that are not part of equals ... worth it ???

PR #772, opened 2019-12-02
TSK-967: Extract module taskana-data to distribute data for tests and…

benjamineckstein:
… examples

Part two of merging taskana-rest-spring and taskana-rest-spring-test ... we need to extract the data module to reuse it. 

The next step will be to merge the two modules.


lib/taskana-data/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+			<groupId>org.mybatis</groupId>
+			<artifactId>mybatis</artifactId>
+			<version>${version.mybatis}</version>
+		</dependency>
+
+		<dependency>
+			<groupId>org.slf4j</groupId>
+			<artifactId>slf4j-api</artifactId>
+			<version>${version.slf4j}</version>
+		</dependency>
```
mustaphazorgati: please add slf4j-simple to prevent [this](https://stackoverflow.com/questions/7421612/slf4j-failed-to-load-class-org-slf4j-impl-staticloggerbinder) error.

lib/taskana-data/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
```diff
@@ -146,20 +131,36 @@ private StringReader selectSchemaScript(String dbProductName, String schemaName)
...
+    }
+
+    static BufferedReader getScriptBufferedStream(String script) {
+        return Optional.ofNullable(SampleDataGenerator.class.getResourceAsStream(script)).map(
+            inputStream -> new BufferedReader(
+                new InputStreamReader(inputStream, StandardCharsets.UTF_8))).orElse(null);
+    }
+
+    static ArrayList<String> getDefaultScripts() {
+        String[] script = {WORKBASKET, DISTRIBUTION_TARGETS, CLASSIFICATION, TASK, ATTACHMENT, WORKBASKET_ACCESS_LIST,
```
mustaphazorgati: Maybe use a stream which generates all Data instead of manually doing it?

lib/taskana-data/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
```diff
@@ -146,20 +131,36 @@ private StringReader selectSchemaScript(String dbProductName, String schemaName)
...
+            getScriptBufferedStream(script).lines().collect(Collectors.joining(System.lineSeparator())));
+    }
+
+    static BufferedReader getScriptBufferedStream(String script) {
+        return Optional.ofNullable(SampleDataGenerator.class.getResourceAsStream(script)).map(
+            inputStream -> new BufferedReader(
+                new InputStreamReader(inputStream, StandardCharsets.UTF_8))).orElse(null);
+    }
+
+    static ArrayList<String> getDefaultScripts() {
```
mustaphazorgati: Please return List<String>

lib/taskana-data/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
```diff
@@ -146,20 +131,36 @@ private StringReader selectSchemaScript(String dbProductName, String schemaName)
...
+            runner.runScript(getScriptBufferedStream(CHECK_HISTORY_EVENT_EXIST));
+            runner.runScript(getScriptBufferedStream(CLEAR_HISTORY_EVENTS));
             scriptsList.add(HISTORY_EVENT);
         } catch (Exception e) {
             LOGGER.error("The HISTORY_EVENTS table is not created");
         }
         return scriptsList.toArray(new String[0]);
     }
+
+    static String parseAndReplace(LocalDateTime now, String script) {
```
mustaphazorgati: This can be private again (since you moved the test)

rest/taskana-rest-spring-example-common/pom.xml
```diff
@@ -14,13 +14,28 @@
 		<version>1.1.6-SNAPSHOT</version>
 		<relativePath>../pom.xml</relativePath>
 	</parent>
-
+	<dependencyManagement>
```
mustaphazorgati: Why did you add this? Personally I do not like using dependencyManagement unless it is necessary

rest/taskana-rest-spring/pom.xml
```diff
@@ -14,6 +14,18 @@
 		<relativePath>../pom.xml</relativePath>
 	</parent>
 
+	<dependencyManagement>
```
mustaphazorgati: As above: Is there any reason for this?

rest/taskana-rest-spring/src/test/java/pro/taskana/rest/security/SampleLoginModule.java
```diff
@@ -0,0 +1,84 @@
...
+
+import org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter;
+
+import pro.taskana.ldap.LdapCacheTestImpl;
+import pro.taskana.rest.resource.AccessIdResource;
+import pro.taskana.security.GroupPrincipal;
+import pro.taskana.security.UserPrincipal;
+
+/**
+ * TODO.
```
mustaphazorgati: Please make a real comment.

lib/taskana-data/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
```diff
@@ -146,20 +131,36 @@ private StringReader selectSchemaScript(String dbProductName, String schemaName)
...
+                new InputStreamReader(inputStream, StandardCharsets.UTF_8))).orElse(null);
+    }
+
+    static ArrayList<String> getDefaultScripts() {
+        String[] script = {WORKBASKET, DISTRIBUTION_TARGETS, CLASSIFICATION, TASK, ATTACHMENT, WORKBASKET_ACCESS_LIST,
+            OBJECT_REFERENCE};
+        return new ArrayList<>(Arrays.asList(script));
+    }
+
+    private static boolean isPostgreSQL(String databaseProductName) {
```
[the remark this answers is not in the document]
benjamineckstein: 3 lines of code it not that much magic to justify including the taskana core dependency

lib/taskana-data/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
```diff
@@ -125,18 +120,8 @@ private static String replaceRelativeTimeFunction(LocalDateTime now, String sql)
...
-            return replaceRelativeTimeFunction(now,
-                bufferedReader.lines().collect(Collectors.joining(System.lineSeparator())));
-        } catch (IOException e) {
-            throw new RuntimeException(e);
-        }
-    }
-
     private StringReader selectSchemaScript(String dbProductName, String schemaName) {
-        return new StringReader(TaskanaEngineImpl.isPostgreSQL(dbProductName)
+        return new StringReader(isPostgreSQL(dbProductName)
```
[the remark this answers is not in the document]
benjamineckstein: Not available. No core dependency. See above

lib/taskana-data/pom.xml
benjamineckstein: thx will fix this.

lib/taskana-data/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
benjamineckstein: if we going more into refactoring. let's return streams.

rest/taskana-rest-spring/src/test/java/pro/taskana/rest/security/SampleLoginModule.java
benjamineckstein: Haha .... good catch. Copied from the common module.

lib/taskana-data/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
benjamineckstein: i see .. we are on the same side ;)

lib/taskana-data/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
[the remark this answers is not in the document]
mustaphazorgati: oh. yes. forgot that :man_facepalming:

lib/taskana-data/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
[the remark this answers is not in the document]
mustaphazorgati: yep.

lib/taskana-data/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
mustaphazorgati: i like that

lib/taskana-data/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
```diff
@@ -146,20 +131,36 @@ private StringReader selectSchemaScript(String dbProductName, String schemaName)
...
+
+    static String parseAndReplace(LocalDateTime now, String script) {
+        return replaceRelativeTimeFunction(now,
+            getScriptBufferedStream(script).lines().collect(Collectors.joining(System.lineSeparator())));
+    }
+
+    static BufferedReader getScriptBufferedStream(String script) {
+        return Optional.ofNullable(SampleDataGenerator.class.getResourceAsStream(script)).map(
+            inputStream -> new BufferedReader(
+                new InputStreamReader(inputStream, StandardCharsets.UTF_8))).orElse(null);
```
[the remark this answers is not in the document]
benjamineckstein: IS this needed? What exception. Checked or unchecked?
 
Default behavior is returning null 
image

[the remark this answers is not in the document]
mustaphazorgati: The idea was a RuntimeException which says that the given file does not exist. But having a second thought about this our unit test will already tell us if a file does not exist.. (this should be impossible)

given your example I'd agree with you - null is perfectly fine

lib/pom.xml
```diff
@@ -16,6 +16,7 @@
 
 	<modules>
 		<module>taskana-core</module>
+		<module>taskana-data</module>
```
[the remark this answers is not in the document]
benjamineckstein: done

lib/taskana-data/pom.xml
benjamineckstein: done

lib/taskana-data/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
benjamineckstein: done

lib/taskana-data/pom.xml
benjamineckstein: It is worth noting that embedded components such as libraries or frameworks should never declare a dependency on any SLF4J binding.  This is because when a library declares a compile-time dependency on an SLF4J binding, it imposes that binding on the end-user. Obviously, this negates SLF4J's basic purpose. Consequently, they should only depend on the slf4j-api library.

https://www.baeldung.com/slf4j-classpath-multiple-bindings

benjamineckstein: this creates conflicts with other modules (rest-spring)

benjamineckstein: added simple for the test scope

rest/taskana-rest-spring/pom.xml
benjamineckstein: lets postpone this to another ticket. tried to get rid of it.

rest/taskana-rest-spring-example-common/pom.xml
benjamineckstein: lets postpone this to another ticket. tried to get rid of it.

lib/taskana-data/pom.xml
mustaphazorgati: I am a little confused now. Maybe talk later / tomorrow briefly about it?

lib/taskana-data/src/test/java/pro/taskana/sampledata/SampleDataGeneratorTest.java
```diff
@@ -0,0 +1,26 @@
...
+    void getScriptsNotEmpty() {
+        Stream<String> scripts = SampleDataGenerator.getDefaultScripts();
+        Assertions.assertNotNull(scripts);
+        Assertions.assertTrue(scripts.count() > 0);
+    }
+
+    @Test
+    void getScriptsFileExists() {
+        Stream<String> scripts = SampleDataGenerator.getDefaultScripts();
+        scripts.forEach(s -> Assertions.assertNotNull(SampleDataGenerator.getScriptBufferedStream(s)));
```
mustaphazorgati: SampleDataGenerator.getDefaultScripts()
.map(SampleDataGenerator::getScriptBufferedStream)
.forEach(Assertions::assertNotNull))

mustaphazorgati: what do you think of that? I <3 method references :) makes it look so "clean"

rest/taskana-rest-spring-example-common/pom.xml
mustaphazorgati: please create it :)

lib/taskana-data/src/test/java/pro/taskana/sampledata/SampleDataGeneratorTest.java
benjamineckstein: done

PR #777, opened 2019-12-09
TSK-975: Remove unnecessary initializers

BVier:


BVier:
TSK-977 is based upon this task.
I splited them because TSK-977 is about the enforcing of linting, not the correction of older linting errors

BVier:
But in the end you're right - if you approve TSK-977 this PR is redundant.

benjamineckstein:
so please close one of these two prs. I will review the other one again then. Also please mark discussions as resolved if this point is no longer valid.

benjamineckstein:
included in TSK 977 PR

PR #783, opened 2019-12-12
TSK-909

gitgoodjhe:
-Added a new CallbackState which is needed when a task was claimed by TASKANA
-Added ruleset logic to determine if the public API is allowed to set the desired CallbackState
-Added test for the ruleset logic

### DIFF
--- a/lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
+++ b/lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
@@ -459,7 +459,7 @@ public class TaskServiceImpl implements TaskService {
 
             Iterator<String> taskIdIterator = externalIds.iterator();
             while (taskIdIterator.hasNext()) {
-                removeSingleTaskForCallbackStateByExternalId(bulkLog, taskSummaries, taskIdIterator);
+                removeSingleTaskForCallbackStateByExternalId(bulkLog, taskSummaries, taskIdIterator, state);
             }
             if (!externalIds.isEmpty()) {
                 taskMapper.setCallbackStateMultiple(externalIds, state);
@@ -505,7 +505,7 @@ public class TaskServiceImpl implements TaskService {
 
     private void removeSingleTaskForCallbackStateByExternalId(BulkOperationResults<String,
         TaskanaException> bulkLog,
-        List<MinimalTaskSummary> taskSummaries, Iterator<String> externalIdIterator) {
+        List<MinimalTaskSummary> taskSummaries, Iterator<String> externalIdIterator, CallbackState desiredCallbackState) {
         LOGGER.debug("entry to removeSingleTask()");
         String currentExternalId = externalIdIterator.next();
         if (currentExternalId == null || currentExternalId.equals("")) {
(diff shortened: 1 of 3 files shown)


lib/taskana-core/src/test/java/acceptance/task/CallbackStateAccTest.java
```diff
@@ -155,13 +158,82 @@ void testUpdateOfCallbackState()
...
+
+        //It's only allowed to set the CallbackState to REQUIRED if the TaskState doesn't equal COMPLETE
+        //Therefore 1 task should not get updated
+        BulkOperationResults<String, TaskanaException> bulkResult4 = taskService.setCallbackStateForTasks(externalIds,
+            CallbackState.CALLBACK_PROCESSING_REQUIRED);
+        assertTrue(bulkResult4.containsErrors());
+        List<String> failedTaskIds4 = bulkResult4.getFailedIds();
+        assertTrue(failedTaskIds4.size() == 1);
+
+    }
```
benjamineckstein: Test to loooooonnngg ... can you split them up into multiple tests?

lib/taskana-core/src/test/java/acceptance/task/CallbackStateAccTest.java
```diff
@@ -207,7 +281,9 @@ void testQueriesWithCallbackState()
 
     }
 
-    private TaskImpl createTask(TaskService taskService, CallbackState callbackState) throws WorkbasketNotFoundException, ClassificationNotFoundException, NotAuthorizedException, TaskAlreadyExistException, InvalidArgumentException {
+    private TaskImpl createTask(TaskService taskService, CallbackState callbackState)
```
benjamineckstein: unneeded formatting? Also on other files too. It does not matter, but makes your PR bigger and thus takes longer to review.

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
```diff
@@ -510,14 +510,50 @@ private void removeSingleTaskForCallbackStateByExternalId(BulkOperationResults<S
...
+            } else if (!desiredCallbackStateCanBeSetForFoundSummary(foundSummary, desiredCallbackState)) {
                 bulkLog.addError(currentExternalId, new InvalidStateException(currentExternalId));
                 externalIdIterator.remove();
             }
+
         }
         LOGGER.debug("exit from removeSingleTask()");
     }
 
+    private boolean desiredCallbackStateCanBeSetForFoundSummary(MinimalTaskSummary foundSummary, CallbackState desiredCallbackState) {
```
benjamineckstein: this is such a good case of unit testing domain logic. please consider making this package-private and adding a few tests to verify this logic.

lib/taskana-core/src/test/java/acceptance/task/CallbackStateAccTest.java
gitgoodjhe: Agreed, this was auto-formatting (using our QA IntelliJ Formatter).

lib/taskana-core/src/test/java/acceptance/task/CallbackStateAccTest.java
gitgoodjhe: Done

lib/taskana-core/src/main/java/pro/taskana/impl/TaskServiceImpl.java
gitgoodjhe: The acceptance test includes full code coverage of this logic. Since there will be a refactoring of the TaskServiceImpl class very soon, any unit tests would have to be adjusted again, while the acceptance tests stays as it is. For these reasons I won't add extra unit tests for now.

PR #789, opened 2019-12-19
TSK-976: Set up eslint

BVier:


web/.eslintrc.js
```diff
@@ -0,0 +1,105 @@
...
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "project": "tsconfig.json",
+  },
+  "plugins": [
+        "@typescript-eslint",
+        "@typescript-eslint/tslint"
+  ],
+  "rules": {
+    "arrow-parens": ["off", "as-needed"],
```
mustaphazorgati: where did you get the rules from? isn't the airbnb base good enough? Or is this just to deactivate the rules for now?

web/package.json
```diff
@@ -10,11 +10,12 @@
     "build:prod-silent": "ng build --prod=true --no-progress",
     "test": "ng test --karma-config karma.conf.js --watch=false --browsers Firefox",
     "test:watch": "ng test --karma-config karma.conf.js --browsers Chrome",
-    "lint": "ng lint"
+    "lint": "eslint src/app/**/*.ts"
   },
   "private": true,
   "dependencies": {
     "@angular/animations": "8.2.11",
+    "@angular/core": "8.2.11",
```
mustaphazorgati: please use ^8.2.11

web/package.json
```diff
@@ -64,7 +71,7 @@
     "protractor": "5.4.0",
     "ts-mockito": "2.3.0",
     "ts-node": "4.1.0",
-    "tslint": "5.9.1",
+    "tslint": "^5.9.1",
```
mustaphazorgati: still necessary, even after removal?

web/.eslintrc.js
BVier: Must rules are just deactivated to make the build run until they are fixed.
Some rules didn't matched our format but could be configured to match it.
(e.g. indent is 2 not 4)

web/package.json
BVier: ok.
Only for angular cor?

web/package.json
BVier: It shouldn't be, but my console required it...
Maybe it's because of the 'ng lint' settings which I didn't change yet.

BVier: But pls try it without this dependency - maybe it's only a local problem of mine!

web/package.json
BVier: ✔️

web/package.json
BVier: I thought so, but I had problems when I linted with my cmd.
After the rebase it works without this dependecy so I think it's ok to remove it!

BVier:
Sry, I have no idea why the build is failing...

mustaphazorgati:
> Sry, I have no idea why the build is failing...
Probably node cache. Restarted build. Let's see

mustaphazorgati:
Almost.. Malformed package-lock.json

mustaphazorgati:
fount it: https://github.com/angular/angular-cli/issues/15493#issuecomment-529098088

web/package.json
mustaphazorgati: It's necessary for the eslint-tslint plugin

PR #790, opened 2019-12-24
TSK-1000: classifications are now being shown in the UI

mustaphazorgati:


web/src/app/shared/services/classifications/classifications.service.ts
```diff
@@ -97,27 +108,15 @@ export class ClassificationsService {
...
+    const classificationTypes: Observable<string> = this.classificationCategoriesService.getSelectedClassificationType();
     return combineLatest(
       [classificationRef,
         classificationTypes]
     ).pipe(
       map(
-        (classification: any, classificationType: any) => {
-          return classification.classifications ? this.buildHierarchy(classification.classifications, classificationType) : [];
+        (classification: any[]) => {
+          return classification[0].classifications ? this.buildHierarchy(classification[0].classifications, classification[1]) : [];
```
benjamineckstein: what if [0] does not exist? At least it looks like a code smell.

mustaphazorgati: since we are passing an array with 2 values in the compineLatest() function we always have a length of 2.
Honestly I don't like the change either but I didn't want to invest too much time in looking for alternatives. combineLatest without an array is depreceated. Therefore @BVier switched to an array (which is recommended)

BVier: Yes it is absolutely a code smell, but this is the best functional hotfix.
In my opinion we should refactor the whole getter method in a seperate ticket.
After this the problem should be resolved.

PR #842, opened 2020-01-14
Tsk 1021

BerndBreier:


lib/taskana-core/src/main/java/pro/taskana/mappings/WorkbasketMapper.java
```diff
@@ -25,8 +27,8 @@
       value = {
         @Result(property = "id", column = "ID"),
         @Result(property = "key", column = "KEY"),
-        @Result(property = "created", column = "CREATED"),
-        @Result(property = "modified", column = "MODIFIED"),
+        @Result(property = "created", column = "CREATED", jdbcType = JdbcType.TIMESTAMP, javaType = java.time.Instant.class, typeHandler = InstantTypeHandler.class),
```
benjamineckstein: i dont like that. too much risk of forgetting any "timestamp" field. Think about new ones in the future added by someone else ...

BerndBreier: That's correct. But do you have a better alternative?

mustaphazorgati:
Hey @BerndBreier,

looks like something went wrong. You got all your commits twice and you have a merge within the same branch. Let's check this out together tomorrow.

mustaphazorgati:
And also remove the changes to the package-lock file

BerndBreier:
Please don't merge this PR. We are currently trying to convince Mybatis to use the statements from our InstantTypeHandler in their InstantTypeHandler. This would simplify the code enormously because no annotation would need to be changed ...
Only if Mybatis refuses, we should use the solution proposed by this PR

BerndBreier:
Found a better way to fix the problem. If the type handler is registered >>before<< the mappers are registered, it is used automatically and no annotations are required

PR #843, opened 2020-01-14
TSK-998: Fix simple linting errors

BVier:
Only fixes which are done by replacing

### DIFF
--- a/web/src/app/administration/classification/details/classification-details.component.ts
+++ b/web/src/app/administration/classification/details/classification-details.component.ts
@@ -18,9 +18,7 @@ import { AlertService } from 'app/services/alert/alert.service';
 import { TreeService } from 'app/services/tree/tree.service';
 import { RemoveConfirmationService } from 'app/services/remove-confirmation/remove-confirmation.service';
 
-// tslint:disable:max-line-length
 import { ClassificationCategoriesService } from 'app/shared/services/classifications/classification-categories.service';
-// tslint:enable:max-line-length
 import { DomainService } from 'app/services/domain/domain.service';
 import { Pair } from 'app/models/pair';
 import { NgForm } from '@angular/forms';
@@ -112,7 +110,7 @@ export class ClassificationDetailsComponent implements OnInit, OnDestroy {
         this.fillClassificationInformation(this.classification ? this.classification : new ClassificationDefinition());
       }
 
-      if (!this.classification || this.classification.classificationId !== id && id) {
+      if (!this.classification || (this.classification.classificationId !== id && id)) {
         this.selectClassification(id);
       }
     });
@@ -128,7 +126,7 @@ export class ClassificationDetailsComponent implements OnInit, OnDestroy {
         // TSK-891 fix: The property is already set and is crucial value
         // Wrapped with an if to set a default if not already set.
         if (!this.classification.category) {
-          this.classification.category = categories[0];
+          [this.classification.category] = categories;
         }
       }
     });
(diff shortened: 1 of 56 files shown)


web/src/app/administration/classification/details/classification-details.component.ts
```diff
@@ -128,7 +126,7 @@ export class ClassificationDetailsComponent implements OnInit, OnDestroy {
         // TSK-891 fix: The property is already set and is crucial value
         // Wrapped with an if to set a default if not already set.
         if (!this.classification.category) {
-          this.classification.category = categories[0];
+          [this.classification.category] = categories;
```
benjamineckstein: haha destruct operator for arrays ... very tricky thing for non js devs ;)

web/src/app/administration/classification/details/classification-details.component.ts
```diff
@@ -224,18 +222,17 @@ export class ClassificationDetailsComponent implements OnInit, OnDestroy {
...
 
   private async selectClassification(id: string) {
-    if (this.classificationIsAlreadySelected()) {
-      return true;
+    if (!this.classificationIsAlreadySelected()) {
+      this.requestInProgress = true;
+      const classification = await this.classificationsService.getClassification(id);
+      this.fillClassificationInformation(classification);
+      this.classificationsService.selectClassification(classification);
+      this.requestInProgress = false;
```
benjamineckstein: return true is missing? not  needed?

web/src/app/administration/classification/details/classification-details.component.ts
```diff
@@ -224,18 +222,17 @@ export class ClassificationDetailsComponent implements OnInit, OnDestroy {
...
-    this.requestInProgress = true;
-    const classification = await this.classificationsService.getClassification(id);
-    this.fillClassificationInformation(classification);
-    this.classificationsService.selectClassification(classification);
-    this.requestInProgress = false;
   }
 
   private classificationIsAlreadySelected(): boolean {
-    if (this.action === ACTION.CREATE && this.classification) { return true; }
+    return this.action === ACTION.CREATE && !!this.classification;
```
benjamineckstein: returns true & false now. intended?

web/src/app/history/task-query/task-query.component.ts
```diff
@@ -184,13 +183,8 @@ export class TaskQueryComponent implements OnInit {
       false
     ).subscribe(taskQueryResource => {
       this.requestInProgressService.setRequestInProgress(false);
-      if (!taskQueryResource.taskHistoryEvents) {
```
benjamineckstein: if case not needed??

web/src/app/administration/classification/details/classification-details.component.ts
BVier: Yes, it is - i googled twice! ;)

web/src/app/administration/classification/details/classification-details.component.ts
BVier: It's definitely intended because a function should always return a value fro the same type - in this case boolean.

web/src/app/history/task-query/task-query.component.ts
BVier: I inserted the if case by inserting ' = taskQueryResource.taskHistoryEvents ? ...'

So we get the same behavior in two lines and without an empty return statement

web/src/app/administration/classification/details/classification-details.component.ts
BVier: No, because the return value is never catched.

PR #852, opened 2020-01-20
Tsk1029 Deletion of workbasket as business admin not possible without explicit permissions

benjamineckstein:


lib/taskana-core/src/main/java/pro/taskana/security/CurrentUserContext.java
```diff
@@ -77,6 +79,31 @@ public static String getUserid() {
...
+    Subject subject = Subject.getSubject(AccessController.getContext());
+    if (subject == null) {
+      // dont add authorisation if none is available.
+      return supplier.get();
+    }
+    Set<Principal> principals = subject.getPrincipals();
+    Set<Object> privateCredentials = subject.getPrivateCredentials();
+    Set<Object> publicCredentials = subject.getPublicCredentials();
+
+    principals.add(new GroupPrincipal("admin"));
```
holgerhagen: This only works if you do have a user called "admin" configured as admin in taskana.properties, doesn't it? I guess we have to think about this a bit more.

holgerhagen: Don't we change the current Subject as well, if we do it this way? From my perspective it might be better to create a dedicated Subject from scratch instead of patching the current subject.

rest/taskana-rest-spring/src/test/java/pro/taskana/rest/WorkbasketControllerIntTest.java
```diff
@@ -107,6 +108,25 @@ void testGetSecondPageSortedByKey() {
     assertNotNull(response.getBody().getLink(Link.REL_PREVIOUS));
   }
 
+  /**
+   * Bug Ticket TSK-1029.
+   * Businessadmin is allowed to delete any workbasket ticket without user related access
+   * restrictions.
+   */
+  @Test
+  void testDeleteWorkbasketPermissionWithBusinessAdmin() {
```
holgerhagen: testDeleteWorkbasketAsBusinessAdminWithoutExplicitReadPermission?

lib/taskana-core/src/main/java/pro/taskana/security/CurrentUserContext.java
BerndBreier: we could reuse method JobScheduler.getAdminSubject() that creates an admin subject.
May be, the method should be factored out, e.g. moved to InternalTaskanaEngine ...

rest/taskana-rest-spring/src/test/java/pro/taskana/rest/WorkbasketControllerIntTest.java
benjamineckstein: Done

lib/taskana-core/src/main/java/pro/taskana/security/CurrentUserContext.java
benjamineckstein: I fixed that issue and added a test to verify that this works as intended

PR #857, opened 2020-01-22
TSK-1021 persist timestamps in UTC representation

BerndBreier:


lib/taskana-core/src/main/java/pro/taskana/impl/TaskanaEngineImpl.java
```diff
@@ -232,6 +234,9 @@ protected SqlSessionManager createSqlSessionManager() {
           e.getCause());
     }
 
+    // register type handlers
+    configuration.getTypeHandlerRegistry().register(MapTypeHandler.class);
+    configuration.getTypeHandlerRegistry().register(Instant.class, InstantTypeHandler.class);
```
mustaphazorgati: Just out of curiosity: 
Should we instantiate the Handlers, so that mybatis doesnt do it via reflections?

mustaphazorgati: Does that even matter?

lib/taskana-core/src/test/java/acceptance/persistence/UpdateObjectsUseUtcTimeStampsAccTest.java
```diff
@@ -0,0 +1,243 @@
...
+import pro.taskana.exceptions.WorkbasketNotFoundException;
+import pro.taskana.impl.JobServiceImpl;
+import pro.taskana.impl.TaskImpl;
+import pro.taskana.jobs.ScheduledJob;
+import pro.taskana.security.JaasExtension;
+import pro.taskana.security.WithAccessId;
+
+/** Acceptance test for access to timestamps from different timezones. */
+@ExtendWith(JaasExtension.class)
+public class UpdateObjectsUseUtcTimeStampsAccTest extends AbstractAccTest {
```
mustaphazorgati: Since the timestamp behaviour should be done globally, do we need specific (complicated) test cases like this?
Can't we just do it with one entity?

lib/taskana-data/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
```diff
@@ -33,18 +35,18 @@
...
+  private final ZonedDateTime now;
   /**
    * This value cannot be automatically obtained by connection.getSchema(), because setting not yet
    * existing schema will result into an SQL Exception.
    */
   private final String schema;
 
   public SampleDataGenerator(DataSource dataSource, String schema) {
-    this(dataSource, schema, LocalDateTime.now());
+    this(dataSource, schema, Instant.now().atZone(ZoneId.of("UTC")));
```
mustaphazorgati: ZonedDateTime.now()

BerndBreier: ZonedDateTime.now() gives me the actual time in the current time zone, not in the UTC time zone. If I invoke ZonedDateTime.now() and print it, I see 2020-01-22 13:53:05, which is the current moment in the local time zone.
If I print at the same time the output of Instant.now().atZone(ZoneId.of("UTC")), I see 2020-01-22 12:53:05 which is the current moment in UTC time zone.

lib/taskana-core/src/test/java/acceptance/persistence/UpdateObjectsUseUtcTimeStampsAccTest.java
BerndBreier: This testcase was created when we had to annotate each individual timestamp field.
However, I think I doesn't hurt to test the timestamps on all persistent objects...

lib/taskana-core/src/main/java/pro/taskana/impl/TaskanaEngineImpl.java
BerndBreier: I even don't know an interface that we could use to pass instantiated handlers to mybatis...

lib/taskana-core/src/test/java/acceptance/persistence/UpdateObjectsUseUtcTimeStampsAccTest.java
mustaphazorgati: ok

lib/taskana-core/src/main/java/pro/taskana/impl/TaskanaEngineImpl.java
mustaphazorgati: https://mybatis.org/mybatis-3/apidocs/reference/org/apache/ibatis/type/TypeHandlerRegistry.html

would be
register(TypeHandler<T> typeHandler) for the MapTypeHandler
and 
register(Class<T> javaType, TypeHandler<? extends T> typeHandler) for the InstantTypeHandler

mustaphazorgati: looking at the sourcecode of mybatis the register(Class<T> ..) methods do some reflection magic and then call the regular register(TypeHandler<?> ..) methods. Do we want that reflection magic to happen?

BerndBreier: I'll see whether this works also and if so, register the instantiated typehandlers ...

lib/taskana-data/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
mustaphazorgati: okay. Why did you use ZonedDateTime.now() in the SqlReplacerTests then? :)

mustaphazorgati: I was a little confused, because the change from LocalDateTime -> ZonedDateTime was not consistent. That's why I am asking :+1:

BerndBreier: In the SqlReplacerTests, it doesn't matter if you are using local time or utc time. These times are not persisted, only the correct matching and correct relative calculations are checked. However, I can change this in the SqlReplacer, too.

BerndBreier: wanted to say " ... change this in the SqlReplacerTests, too"

mustaphazorgati: If it doesn't matter then I dont care.

lib/taskana-core/src/main/java/pro/taskana/impl/TaskanaEngineImpl.java
BerndBreier: Works fine. I switched to the registration with instantiated handlers ...

PR #860, opened 2020-01-22
TSK-987: Fix some problems reported by spotbugs

benjamineckstein:


history/taskana-simplehistory-provider/src/main/java/pro/taskana/simplehistory/configuration/DbSchemaCreator.java
```diff
@@ -44,12 +44,12 @@ public void run() throws SQLException {
       runner.setStopOnError(true);
       runner.setLogWriter(logWriter);
       runner.setErrorLogWriter(errorLogWriter);
-      
-      InputStream resourceAsStream = this.getClass().getResourceAsStream(DB_SCHEMA);
+
+      InputStream resourceAsStream = DbSchemaCreator.class.getResourceAsStream(DB_SCHEMA);
```
mustaphazorgati: why is this a bug? Performance? Since getClass() requires an instant and calls a virtual method?

gitgoodjhe: Probably this: 

UI: Usage of GetResource may be unsafe if class is extended (UI_INHERITANCE_UNSAFE_GETRESOURCE)

Calling this.getClass().getResource(...) could give results other than expected if this class is extended by a class in another package.

Also see this: https://bz.apache.org/netbeans/show_bug.cgi?id=122477

benjamineckstein:
**Before** 

image

**After**

image

PR #864, opened 2020-01-23
TSK-1037: WBDefinitonController accepts existing distributionTargets

BVier:
Additionally allow a correct update of an workbasketDefinitonResource with existing Key-Domain value.

### DIFF
--- a/rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketDefinitionController.java
+++ b/rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketDefinitionController.java
@@ -132,25 +132,26 @@ public class WorkbasketDefinitionController {
     // STEP 1: update or create workbaskets from the import
     for (WorkbasketDefinitionResource definition : definitions) {
       Workbasket importedWb = workbasketDefinitionAssembler.toModel(definition.getWorkbasket());
-      Workbasket workbasket;
+      String newId;
+      Workbasket wbWithoutId = removeId(importedWb);
       if (systemIds.containsKey(logicalId(importedWb))) {
-        workbasket = workbasketService.updateWorkbasket(importedWb);
+        workbasketService.updateWorkbasket(wbWithoutId);
+        newId = systemIds.get(logicalId(importedWb));
       } else {
-        Workbasket wbWithoutId = removeId(importedWb);
-        workbasket = workbasketService.createWorkbasket(wbWithoutId);
+        newId = workbasketService.createWorkbasket(wbWithoutId).getId();
       }
 
       // Since we would have a n² runtime when doing a lookup and updating the access items we
       // decided to
       // simply delete all existing accessItems and create new ones.
       for (WorkbasketAccessItem accessItem :
-          workbasketService.getWorkbasketAccessItems(workbasket.getId())) {
+          workbasketService.getWorkbasketAccessItems(newId)) {
         workbasketService.deleteWorkbasketAccessItem(accessItem.getId());
       }
       for (WorkbasketAccessItem authorization : definition.getAuthorizations()) {
         workbasketService.createWorkbasketAccessItem(authorization);
       }
-      idConversion.put(importedWb.getId(), workbasket.getId());
+      idConversion.put(importedWb.getId(), newId);
     }
 
     // STEP 2: update distribution targets
@@ -160,6 +161,8 @@ public class WorkbasketDefinitionController {
       for (String oldId : definition.getDistributionTargets()) {
         if (idConversion.containsKey(oldId)) {
           distributionTargets.add(idConversion.get(oldId));
+        } else if (systemIds.containsValue(oldId)) {
+          distributionTargets.add(oldId);
         } else {
           throw new InvalidWorkbasketException(
               String.format(
--- a/rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketDefinitionController.java
+++ b/rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketDefinitionController.java
@@ -36,6 +36,7 @@ import pro.taskana.exceptions.NotAuthorizedException;
 import pro.taskana.exceptions.WorkbasketAccessItemAlreadyExistException;
 import pro.taskana.exceptions.WorkbasketAlreadyExistException;
 import pro.taskana.exceptions.WorkbasketNotFoundException;
+import pro.taskana.impl.WorkbasketAccessItemImpl;
 import pro.taskana.rest.resource.WorkbasketDefinitionResource;
 import pro.taskana.rest.resource.WorkbasketDefinitionResourceAssembler;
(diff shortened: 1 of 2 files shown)


rest/taskana-rest-spring/src/main/java/pro/taskana/rest/WorkbasketDefinitionController.java
```diff
@@ -129,25 +129,26 @@
     // STEP 1: update or create workbaskets from the import
     for (WorkbasketDefinitionResource definition : definitions) {
       Workbasket importedWb = workbasketDefinitionAssembler.toModel(definition.getWorkbasket());
-      Workbasket workbasket;
+      String newId;
+      Workbasket wbWithoutId = removeId(importedWb);
       if (systemIds.containsKey(logicalId(importedWb))) {
-        workbasket = workbasketService.updateWorkbasket(importedWb);
+        workbasketService.updateWorkbasket(wbWithoutId);
+        newId = systemIds.get(logicalId(importedWb));
```
BVier: I changed this, because the returned workbasket could contain a wrong id.
This happens in the case of an workbasket with existing key-domain-value but wrong id.
In this case the id should also be removed because the backend doesn't check this properly.

rest/taskana-rest-spring/src/test/java/pro/taskana/rest/WorkbasketDefinitionControllerIntTest.java
```diff
@@ -65,12 +71,7 @@ void resetDb() {
 
   @Test
   void testExportWorkbasketFromDomain() {
-    ResponseEntity<List<WorkbasketDefinitionResource>> response =
```
BVier: This piece of code was called that often so I refactored it to an simple method

rest/taskana-rest-spring/src/test/java/pro/taskana/rest/WorkbasketDefinitionControllerIntTest.java
```diff
@@ -96,46 +97,95 @@ void testExportWorkbasketFromDomain() {
...
-            HttpMethod.GET,
-            restHelper.defaultRequest(),
-            ParameterizedTypeReference.forType(List.class));
-
-    List<String> list = new ArrayList<>();
-    ObjectMapper objMapper = new ObjectMapper();
-    list.add(objMapper.writeValueAsString(response.getBody().get(0)));
-    ResponseEntity<Void> responseImport = importRequest(list);
-    assertEquals(HttpStatus.NO_CONTENT, responseImport.getStatusCode());
+  void testImportEveryWorkbasketFromDomainA() throws IOException {
```
BVier: this test is to avoid the problems Holger had - it imports every workbasket in domain A seperately

rest/taskana-rest-spring/src/test/java/pro/taskana/rest/WorkbasketDefinitionControllerIntTest.java
```diff
@@ -96,46 +97,95 @@ void testExportWorkbasketFromDomain() {
...
+    String letMeBeYourDistributionTarget = w.getWorkbasket().workbasketId;
+    WorkbasketDefinitionResource w2 = wbList.get(1);
+    w2.setDistributionTargets(Collections.singleton(letMeBeYourDistributionTarget));
+    importRequest(HttpStatus.NO_CONTENT, objMapper.writeValueAsString(w),
+        objMapper.writeValueAsString(w2));
+
+    w.getWorkbasket().setWorkbasketId("fancyNewId");
+    w2.setDistributionTargets(Collections.singleton("fancyNewId"));
+    importRequest(HttpStatus.NO_CONTENT, objMapper.writeValueAsString(w),
+        objMapper.writeValueAsString(w2));
```
BVier: this is to define a distribution target within the import-file

rest/taskana-rest-spring/src/test/java/pro/taskana/rest/WorkbasketDefinitionControllerIntTest.java
```diff
@@ -175,6 +234,8 @@ void testNoErrorWhenImportWithSameIdButDifferentKeyAndDomain() throws IOExceptio
     HttpEntity<MultiValueMap<String, Object>> requestEntity = new HttpEntity<>(body, headers);
     String serverUrl = restHelper.toUrl(Mapping.URL_WORKBASKETDEFIITIONS);
 
-    return template.postForEntity(serverUrl, requestEntity, Void.class);
+    ResponseEntity<Void> responseImport = template
+                                              .postForEntity(serverUrl, requestEntity, Void.class);
+    assertEquals(expectedStatus, responseImport.getStatusCode());
```
BVier: this check was always done after the importRequest-function - so I added it internally.

rest/taskana-rest-spring/src/test/java/pro/taskana/rest/WorkbasketDefinitionControllerIntTest.java
benjamineckstein: the imortant part is that a rest call is simulated here. "getXX" is a bit confusing. 
Please rename this to something more meaningful.

rest/taskana-rest-spring/src/test/java/pro/taskana/rest/WorkbasketDefinitionControllerIntTest.java
```diff
@@ -144,27 +194,36 @@ void testFailOnImportDuplicates() throws IOException {
...
-    list.add(objMapper.writeValueAsString(wbDef));
-    ResponseEntity<Void> responseImport = importRequest(list);
-    assertEquals(HttpStatus.NO_CONTENT, responseImport.getStatusCode());
+    List<WorkbasketDefinitionResource> wbList = getExportForDomain("DOMAIN_A").getBody();
+
+    String wbAsItIs = objMapper.writeValueAsString(wbList.get(0));
+    WorkbasketDefinitionResource differentLogicalId = wbList.get(0);
+    differentLogicalId.getWorkbasket().setKey("new Key for this WB");
+
+    // breaks the logic - should we really allow this case?
```
BVier: Should I also fix this test?

When the same Id is given twice to different workbaskets then a link in the distribution targets to this id is ambigious.
(for example: {wb1: ID1, key1, domain1}, {wb2: ID1, key2, domain1}, {wb3: IDx,..., distributionTargets:[ID1]} - which workbasket is the distribution target?)

At the moment the distribution target would only be linked to the last workbasket which uses the id. 
This cannot be the expected behaviour.
We could link the distribution target to all workbaskets with the same Id.

In my opinion it's less confusing when we just deny this case.

holgerhagen: ID and key/domain are both unique identifiers for one workbasket. The case you decribe is invalid. Maybe this was a misunderstanding. Do you have the time to fix this?

PR #865, opened 2020-01-23
TSK-987: Fix Spotbugs - inner class could be made static

benjamineckstein:


lib/taskana-core/src/test/java/acceptance/task/CallbackStateAccTest.java
```diff
@@ -273,7 +273,7 @@ void testInvalidUpdateOfCallbackStateToClaimed()
 
     // claim two tasks
     createdTask1 = (TaskImpl) taskService.forceClaim(createdTask1.getId());
-    createdTask2 = (TaskImpl) taskService.forceClaim(createdTask2.getId());
+    taskService.forceClaim(createdTask2.getId());
```
gitgoodjhe: What is the spotbug error category here? 

Removing this line kinda defeats one aspect of the test, since the state value of createdTask2 will NOT be updated and remains in "READY". The test checks if the state is CLAIMED and the CallbackState is REQUIRED, otherwise it will throw an error and add it to the BulkOperationResults . The test will still pass with this change, because this requirement is still not met if the createdTask2's state is not updated, but it was meant to also check if the error is also thrown when the state is CLAIMED (and CallbackState Claimed as well, which will lead to the error)

benjamineckstein: solved within a zoom session.

PR #869, opened 2020-01-26
TSK-1025: added assertj dependency to every module

mustaphazorgati:


benjamineckstein:
Cool that this is making progress. Can you please add/change at least 1 test with AssertJ Assertions s that Travis proofs that this is working? 
you can use another commit for that ;)

mustaphazorgati:
See the subtask 1053 For that

PR #877, opened 2020-01-27
TSK-989 Filter Tasks by planned TimeIntervals (REST)

gitgoodjhe:
-Added planned param in filter list
-Implemented needed methods
-Added Test

### DIFF
--- a/rest/taskana-rest-spring/src/main/java/pro/taskana/rest/TaskController.java
+++ b/rest/taskana-rest-spring/src/main/java/pro/taskana/rest/TaskController.java
@@ -1,5 +1,6 @@
 package pro.taskana.rest;
 
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
 import org.slf4j.Logger;
@@ -27,6 +28,7 @@ import pro.taskana.TaskQuery;
 import pro.taskana.TaskService;
 import pro.taskana.TaskState;
 import pro.taskana.TaskSummary;
+import pro.taskana.TimeInterval;
 import pro.taskana.exceptions.AttachmentPersistenceException;
 import pro.taskana.exceptions.ClassificationNotFoundException;
 import pro.taskana.exceptions.ConcurrencyException;
@@ -70,10 +72,16 @@ public class TaskController extends AbstractPagingController {
   private static final String POR_SYSTEM = "por.system";
   private static final String POR_COMPANY = "por.company";
   private static final String DUE = "due";
+  private static final String DUE_TO = "due-until";
+  private static final String DUE_FROM = "due-from";
   private static final String PLANNED = "planned";
+  private static final String PLANNED_TO = "planned-until";
+  private static final String PLANNED_FROM = "planned-from";
 
   private static final String SORT_BY = "sort-by";
   private static final String SORT_DIRECTION = "order";
+
+  private static final String INDEFINITE = "";
 
   private TaskService taskService;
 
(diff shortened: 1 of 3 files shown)


rest/taskana-rest-spring/src/main/java/pro/taskana/rest/TaskController.java
```diff
@@ -330,13 +332,65 @@ private TaskQuery applyFilterParams(TaskQuery taskQuery, MultiValueMap<String, S
...
+    TimeInterval[] timeIntervalArray = new TimeInterval[timeIntervalsList.size()];
+
+    TimeInterval[] timeIntervals = timeIntervalsList.toArray(timeIntervalArray);
+
+    return timeIntervals;
+  }
+
+  private TimeInterval determineTimeInterval(String[] instants, int i) {
+
+    if (!instants[i].equals("maxTime") && !instants[i + 1].equals("maxTime")) {
```
benjamineckstein: please consider making maxTime some kind of constant. also is this somewhere documented within the API documentation? this is like a magic string that suddenly does not need to be a valid timestamp / instant object and still have a meaning. 
Personally i would just remove this special cases and find a way to give "null" or only 1 timestamp such a meaning.

rest/taskana-rest-spring/src/test/resources/asciidoc/rest-api.adoc
```diff
@@ -115,6 +115,26 @@ include::{snippets}/GetAllTasksDocTest/response-fields.adoc[]
...
+planned | plannedFrom | plannedTo | due | dueFrom | dueTo +
+
+If it is sufficient to filter the list with a single time interval, use the parameters +
+plannedFrom/dueFrom and plannedTo/plannedTo. +
+Leave one of them out of the request, if you want to have an open interval. +
+
+If it is required to filter the list with multiple planned or due time intervals, the parameters +
+ planned / due can be used as follows instead: +
+
+ - provide a string with an even number of instants separated by a ",". Time intervals will be +
```
mustaphazorgati: provide this example with an example request? Could be better to understand.

rest/taskana-rest-spring/src/test/resources/asciidoc/rest-api.adoc
```diff
@@ -115,6 +115,26 @@ include::{snippets}/GetAllTasksDocTest/response-fields.adoc[]
 The list generated in the response can be filtered using following parameters in the uri: +
 name | priority | state | classification.key | workbasket-id | {workbasket-key , domain} | +
 owner | por.company | por.system | por.instance | por.type | por.value +
+planned | plannedFrom | plannedTo | due | dueFrom | dueTo +
+
+If it is sufficient to filter the list with a single time interval, use the parameters +
```
mustaphazorgati: I know what you are trying to say. But it's hard to read. Maybe something like

> planned stands for a specific timestamp. Whenever an interval is of interest use plannedFrom and plannedTo. plannedFrom and plannedTo have to be used together. due works analog.

rest/taskana-rest-spring/src/main/java/pro/taskana/rest/TaskController.java
gitgoodjhe: Done. Also, we now offer 2 methods to filter. If one time interval is sufficient, the parameters "plannedFrom/dueFrom" and "plannedTo/dueTo" can be used. If one of those is left out it is an open time interval (like you suggested). For the other option to use multiple time intervals as filter, the parameter "planned" can be used and given an even amount of instants. For open time intervals an "empty" string can be used.

benjamineckstein: what happens if someone uses both combinations? -> plannedFrom and planned. Or plannedFrom, plannedTo and planned? 

What is the expected & current behaviour? We open a lot of edge cases that need to be defined and handled.

gitgoodjhe: There is a check for invalid parameter combos in the new commit.

PR #885, opened 2020-01-29
TSK-1051 Expose task description via rest

sofie29:


rest/taskana-rest-spring-example-boot/src/main/resources/application.properties
```diff
@@ -20,7 +20,7 @@ taskana.schemaName=TASKANA
 ########spring.jpa.generate-ddl=true
 ########spring.jpa.properties.hibernate.jdbc.lob.non_contextual_creation=true
 ####### property that control rest api security deploy use true for no security.
-devMode=false
+devMode=true
```
holgerhagen: Please do not check the devMode property. This is only meant to be changed locally.

PR #887, opened 2020-01-29
TSK-1062: removed rawtype from MapTypeHandler

mustaphazorgati:


lib/taskana-core/src/main/java/pro/taskana/impl/persistence/MapTypeHandler.java
```diff
@@ -54,15 +55,16 @@ public Map getNullableResult(ResultSet rs, int columnIndex) throws SQLException
   }
 
   @Override
-  public Map getNullableResult(CallableStatement cs, int columnIndex) throws SQLException {
+  public Map<String, Object> getNullableResult(CallableStatement cs, int columnIndex)
```
benjamineckstein: Can we use optional here? 
I always aim to minimize nulls ;)

lib/taskana-core/src/main/java/pro/taskana/impl/persistence/MapTypeHandler.java
```diff
@@ -17,16 +17,16 @@
...
   private static final Logger LOGGER = LoggerFactory.getLogger(MapTypeHandler.class);
 
   @Override
-  public void setNonNullParameter(PreparedStatement ps, int i, Map parameter, JdbcType jdbcType)
+  public void setNonNullParameter(
+      PreparedStatement ps, int i, Map<String, Object> parameter, JdbcType jdbcType)
       throws SQLException {
     if (parameter != null && parameter.size() > 0) {
-      LOGGER.debug("Input-Map before serializing: ", parameter);
+      LOGGER.debug("Input-Map before serializing: {}", parameter);
```
benjamineckstein: depending on map size ... this could be time consuming. is this a problem?

lib/taskana-core/src/main/java/pro/taskana/impl/persistence/MapTypeHandler.java
```diff
@@ -17,16 +17,16 @@
  *
  * @author EH
  */
-@SuppressWarnings("rawtypes")
```
benjamineckstein: every suppressWarning that we can prevent / refactor is a win !! good job ;)

lib/taskana-core/src/main/java/pro/taskana/impl/persistence/MapTypeHandler.java
mustaphazorgati: it's debug. Not production relevant. 
+ As soon as we get slf4j 2.0 we will use the lambda method and thus this will be no issue

lib/taskana-core/src/main/java/pro/taskana/impl/persistence/MapTypeHandler.java
mustaphazorgati: unfortunately we can't. We are extending the Mybatis `TypeHandler` class and are bound to its signature

PR #893, opened 2020-01-31
Tsk 991

BerndBreier:


lib/taskana-core/src/main/java/pro/taskana/report/api/row/TimestampRow.java
```diff
@@ -1,11 +1,12 @@
-package pro.taskana.impl.report.row;
```
mustaphazorgati: this should be internal

lib/taskana-core/src/test/java/pro/taskana/ArchitectureTest.java
```diff
@@ -16,17 +19,34 @@
...
 class ArchitectureTest {
-
   private static JavaClasses importedClasses;
 
+  DescribedPredicate<JavaClass> doNotContain(String... namesToBeExcluded) {
+    return new DescribedPredicate<JavaClass>("should not be checked") {
+      @Override
+      public boolean apply(JavaClass input) {
+        String matchingClassName =
+            Arrays.asList(namesToBeExcluded).stream()
```
mustaphazorgati: Stream.of() :)

lib/taskana-core/src/test/java/pro/taskana/ArchitectureTest.java
```diff
@@ -16,17 +19,34 @@
...
+                .filter(name -> input.getName().contains(name))
```
mustaphazorgati: .filter(input.getName()::contains)

lib/taskana-core/src/main/java/pro/taskana/report/api/row/TimestampRow.java
BerndBreier: TimestampRow is returned by TimestampReport.getRow(). Therefore, I would consider it part of the API - after all it is a datastructure that the client code must understand ...

lib/taskana-core/src/test/java/pro/taskana/ArchitectureTest.java
BerndBreier: Granted, that's better

PR #898, opened 2020-02-04
TSK-991: Add more architecture tests

benjamineckstein:
More Architecture Tests 
Some need to be enabled and resulting errors need to be fixed

### DIFF
--- a/lib/taskana-core/src/test/java/pro/taskana/ArchitectureTest.java
+++ b/lib/taskana-core/src/test/java/pro/taskana/ArchitectureTest.java
@@ -1,16 +1,25 @@
 package pro.taskana;
 
 import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.noClasses;
 import static com.tngtech.archunit.library.GeneralCodingRules.NO_CLASSES_SHOULD_ACCESS_STANDARD_STREAMS;
 import static com.tngtech.archunit.library.GeneralCodingRules.NO_CLASSES_SHOULD_THROW_GENERIC_EXCEPTIONS;
 import static com.tngtech.archunit.library.dependencies.SlicesRuleDefinition.slices;
+import static org.junit.jupiter.api.DynamicTest.dynamicTest;
 
 import com.tngtech.archunit.core.domain.JavaClasses;
 import com.tngtech.archunit.core.importer.ClassFileImporter;
+import com.tngtech.archunit.core.importer.ImportOption.DoNotIncludeTests;
 import com.tngtech.archunit.lang.ArchRule;
+import java.util.Collection;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.DynamicTest;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestFactory;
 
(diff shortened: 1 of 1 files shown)


lib/taskana-core/src/test/java/pro/taskana/ArchitectureTest.java
```diff
@@ -84,20 +83,92 @@ void noClassShouldAccessStandardStreams() {
...
+   * https://www.archunit.org/userguide/html/000_Index.html#_cycle_checks
+   */
+  @TestFactory
+  Collection<DynamicTest> freeOfCycles_subpackages() {
+
+    Stream<String> packagesToTest =
+        Stream.of(
+            "pro.taskana.common.internal.(*)..",
+            "pro.taskana.common.api.(*)..",
+            "pro.taskana.common.internal.(*)..",
```
gitgoodjhe: Why "pro.taskana.common.internal.(*).." twice?

lib/taskana-core/src/test/java/pro/taskana/ArchitectureTest.java
```diff
@@ -84,20 +83,92 @@ void noClassShouldAccessStandardStreams() {
...
+                "..classification..", "..history..", "..report..", "..task..", "..workbasket..");
+    myRule.check(importedClasses);
+  }
+
   @Disabled
-  void freeOfCyclicDependencies() {
-    ArchRule myRule = slices().matching("pro.taskana.(*)..").should()
-                          .notDependOnEachOther();
+  @Test
+  void noClassesShouldDependOnReports() {
```
BerndBreier: Das Problem scheint zu sein, dass der DaysToWorkingDaysConverter, der aktuell im report package liegt, von TaskServiceImpl benutzt wird. Vielleicht sollten wir diesen in das common package refactoren.
Dann könnten wir diesen Test enablen...

lib/taskana-core/src/test/java/pro/taskana/ArchitectureTest.java
```diff
@@ -84,20 +83,92 @@ void noClassShouldAccessStandardStreams() {
...
+            p ->
+                dynamicTest(
+                    p.replaceAll(Pattern.quote("pro.taskana."), "") + " is free of cycles",
+                    () -> slices().matching(p).should().beFreeOfCycles().check(importedClasses)))
+        .collect(Collectors.toList());
+  }
+
+  @Disabled("TBD")
   @Test
+  void commonClassesShouldNotDependOnOtherDomainClasses() {
```
BerndBreier: We could take the result of this test to improve the structure of the product, i.e.
- move xxxCleanupJob classes into the domain packages
- exclude TaskanaEngine/TaskanaEngineImpl from the check
and then enable the test again

lib/taskana-core/src/test/java/pro/taskana/ArchitectureTest.java
```diff
@@ -47,7 +47,7 @@ void apiClassesShouldNotDependOnInternalClasses() {
             .and()
             .haveSimpleNameNotEndingWith("BulkOperationResults")
```
BerndBreier: we could refactor BulkOperationResults into Interface + implementation in order to comply with this test.
Probably it would also be possible to restructure TaskanaHistoryEvent.
In addition, the test is disabled because there are violations in the report.api package. May be, we could refactor these classes to remove the violations

lib/taskana-core/src/test/java/pro/taskana/ArchitectureTest.java
benjamineckstein: safe is safe ... 
jk. done.

PR #906, opened 2020-02-06
TSK-1088 optimistic locking on workbasket update

gitgoodjhe:
- Added optimistic locking for workbasket update
- Adjusted WorkbasketDefinitionController
- Added tests in UpdateWorkbasketAccTest
- Added tests in WorkbasketControllerIntTest

### DIFF
--- a/lib/taskana-core/src/main/java/pro/taskana/workbasket/internal/WorkbasketServiceImpl.java
+++ b/lib/taskana-core/src/main/java/pro/taskana/workbasket/internal/WorkbasketServiceImpl.java
@@ -15,6 +15,7 @@ import org.slf4j.LoggerFactory;
 import pro.taskana.TaskanaEngineConfiguration;
 import pro.taskana.common.api.BulkOperationResults;
 import pro.taskana.common.api.LoggerUtils;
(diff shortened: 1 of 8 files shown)


lib/taskana-core/src/main/java/pro/taskana/workbasket/internal/WorkbasketServiceImpl.java
```diff
@@ -829,6 +849,29 @@ public void deleteWorkbasketAccessItemsForAccessId(String accessId)
...
   }
 
+  /**
+   * Check if current object is based on the newest (by modified).
+   *
+   * @param workbasketImplToUpdate the workbasket to update
+   * @throws ConcurrencyException if the workbasket has been modified by some other process.
+   * @throws WorkbasketNotFoundException if the given workbasket does not exist.
+   */
+  private void checkIfWorkbasketExistsAndVerifyTimestampHasNotChanged(
```
benjamineckstein: please refactor this method to get both Workbaskets as parameters. 
this will make it testable and independent from the database.

gitgoodjhe: Done

PR #926, opened 2020-02-13
TSK-1094 introduce models package for */api and */internal

gitgoodjhe:


lib/taskana-core/src/main/java/pro/taskana/classification/internal/models/ClassificationImpl.java
```diff
@@ -17,7 +17,7 @@
 
   public ClassificationImpl() {}
 
-  ClassificationImpl(ClassificationImpl classification) {
+  public ClassificationImpl(ClassificationImpl classification) {
```
benjamineckstein: why is an impl class public?

lib/taskana-core/src/main/java/pro/taskana/task/internal/models/TaskImpl.java
```diff
@@ -27,7 +27,7 @@
   private CallbackState callbackState;
   private List<Attachment> attachments = new ArrayList<>();
 
-  TaskImpl() {}
+  public TaskImpl() {}
```
benjamineckstein: why public?

lib/taskana-core/src/main/java/pro/taskana/classification/internal/models/ClassificationImpl.java
gitgoodjhe: Because the services create them

lib/taskana-core/src/main/java/pro/taskana/task/internal/models/TaskImpl.java
gitgoodjhe: see above

PR #929, opened 2020-02-13
TSK-1112 Changed @RequestMapping to @PostMapping or @GetMapping

sofie29:


lib/taskana-spring-example/src/main/java/pro/taskana/TaskanaTestController.java
```diff
@@ -19,6 +20,7 @@
 import pro.taskana.workbasket.internal.models.WorkbasketImpl;
 
 /** Rest Controller. */
+@SuppressWarnings("unused")
```
holgerhagen: Do we need this. Can't we simply remove the cause for it?

rest/taskana-rest-spring-example-wildfly/src/main/java/pro/taskana/wildfly/security/LoginErrorController.java
```diff
@@ -3,17 +3,19 @@
...
 /** The loginError controller. */
 @Controller
 public class LoginErrorController {
 
-  @PostMapping
-  @GetMapping
-  @RequestMapping("/loginerror")
-  public RedirectView loginError() {
+  @GetMapping(path = "/loginerror")
+  public RedirectView loginErrorGet() {
```
mustaphazorgati: This is not very nice. Please use @Requestmapping(method = { GET, POST}) instead of duplicating the method. Thanks!

lib/taskana-spring-example/src/main/java/pro/taskana/TaskanaTestController.java
```diff
@@ -100,7 +101,7 @@
   }
 
   @Transactional(rollbackFor = Exception.class)
-  @RequestMapping("/cleanup")
+  @DeleteMapping(path = "/cleanup")
```
mustaphazorgati: Is this intentional? You changed the request method from get to delete.

rest/taskana-rest-spring-example-wildfly/src/main/java/pro/taskana/wildfly/security/LoginErrorController.java
sofie29: This won't fix sonarcloud's issue: "This rule raises an issue when method is missing and when the method parameter is configured with more than one verb"

PR #935, opened 2020-02-17
TSK-1103 Changed vulnerable logging messages

sofie29:


lib/taskana-core/src/main/java/pro/taskana/common/internal/util/LogSanitizer.java
```diff
@@ -0,0 +1,13 @@
+package pro.taskana.common.internal.util;
+
+public class LogSanitizer {
+
+  private LogSanitizer() {
+    throw new IllegalStateException("Utility class");
+  }
+
+  // Remove characters which break the log file pattern. Protects against injection attacks.
+  public static String sanitizeLoggingMessage(String loggingMessage) {
```
mustaphazorgati: make the method generic, so that you don't have to call toString() from the outside?

lib/taskana-core/src/main/java/pro/taskana/common/internal/util/LogSanitizer.java
```diff
@@ -0,0 +1,13 @@
+package pro.taskana.common.internal.util;
+
+public class LogSanitizer {
+
+  private LogSanitizer() {
+    throw new IllegalStateException("Utility class");
+  }
+
+  // Remove characters which break the log file pattern. Protects against injection attacks.
```
mustaphazorgati: Make proper javadoc?

lib/taskana-core/src/main/java/pro/taskana/common/internal/util/LogSanitizer.java
```diff
@@ -0,0 +1,13 @@
+package pro.taskana.common.internal.util;
+
+public class LogSanitizer {
```
mustaphazorgati: Is this used in taskana-core? 
@gitgoodjhe should we move this class to taskana-rest-spring?

gitgoodjhe: Good question, not yet used in core, but might be in the future?

mustaphazorgati: > 
User provided data, such as URL parameters, POST data payloads or cookies, should always be considered untrusted and tainted. Applications logging tainted data could enable an attacker to inject characters that would break the log file pattern. This could be used to block monitors and SIEM (Security Information and Event Management) systems from detecting other malicious events.

This problem could be mitigated by sanitizing the user provided data before logging it.

A dedicated rest issue. Will never happen in core. 
I am for moving this util class to taskana-rest-spring while keeping the same package structure

gitgoodjhe: alright

lib/taskana-core/src/main/java/pro/taskana/common/internal/util/LogSanitizer.java
```diff
@@ -0,0 +1,18 @@
...
+    throw new IllegalStateException("Utility class");
+  }
+
+  /**
+   * Removes characters which break the log file pattern. Protects against injection attacks.
+   *
+   * @param loggingMessage String which should be sanitized
+   * @return sanitized logging message
+   */
+  public static String sanitizeLoggingMessage(Object loggingMessage) {
```
benjamineckstein: class name is already LogSanitizer 
what about we rename this method to stripLineBreakingChars(Object loggingMessage)

mustaphazorgati:
Hey @sofie29!
Did you verify that sonarcloud will find more security issues if you expand the tests (so that a new log message is executed)?

sofie29:
> 
> 
> Hey @sofie29!
> Did you verify that sonarcloud will find more security issues if you expand the tests (so that a new log message is executed)?

Hey @mustaphazorgati, there are logger warnings which are covered by tests but they are not found by sonarcloud.

PR #938, opened 2020-02-20
TSK-1135 Removed JUnit 4 dependency in taskana-simplehistory-spring-test

sofie29:
and converted tests to AssertJ

### DIFF
--- a/history/taskana-simplehistory-spring-test/src/test/java/pro/taskana/TaskHistoryEventControllerIntTest.java
+++ b/history/taskana-simplehistory-spring-test/src/test/java/pro/taskana/TaskHistoryEventControllerIntTest.java
@@ -1,9 +1,7 @@
 package pro.taskana;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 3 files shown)


history/taskana-simplehistory-spring-test/src/test/java/pro/taskana/TaskHistoryEventControllerIntTest.java
```diff
@@ -129,10 +126,10 @@ public void testThrowsExceptionIfInvalidFilterIsUsed() {
           HttpMethod.GET,
           request,
           ParameterizedTypeReference.forType(TaskHistoryEventListResource.class));
-      fail();
+      fail("");
```
gitgoodjhe: Maybe we can add a fitting message to the fail like "Invalid filter is used" instead of an empty string. Also I'm not sure if we need this at all, can't we just use the assertThatThrownBy-pattern that we used in other cases? Pls consult with @benjamineckstein

history/taskana-simplehistory-spring-test/src/test/java/pro/taskana/TaskHistoryEventControllerIntTest.java
```diff
@@ -64,7 +61,7 @@
 
   @Autowired private DataSource dataSource;
 
-  @Before
+  @BeforeEach
   public void before() {
```
gitgoodjhe: Maybe we should name this beforeEach to be consitent with the annotation change

history/taskana-simplehistory-spring-test/src/test/java/pro/taskana/TaskHistoryEventControllerIntTest.java
```diff
@@ -146,10 +143,10 @@ public void testGetHistoryEventOfDate() {
           HttpMethod.GET,
           request,
           ParameterizedTypeReference.forType(TaskHistoryEventListResource.class));
-      fail();
+      fail("");
```
gitgoodjhe: see above

history/taskana-simplehistory-spring-test/src/test/java/pro/taskana/TaskHistoryEventControllerIntTest.java
mustaphazorgati: Yes, we should.

PR #939, opened 2020-02-20
TSK-1131 updateOwner on multiple tasks

BerndBreier:


lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -757,6 +808,77 @@ Duration calculateDuration(
...
+        bulkLog.addError(
+            taskId,
+            new NotAuthorizedException(
+                "Current user not authorized for task " + taskId, currentUserId));
+      }
+    }
+    return tasksAuthorizedFor;
+  }
+
+  private List<String> getAccessIds() {
```
holgerhagen: This would be better placed in TaskanaEngine or even better in CurrentUserContext?

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -617,6 +619,55 @@ public void forceDeleteTask(String taskId)
...
+  public BulkOperationResults<String, TaskanaException> setOwnerOfTasks(
+      String owner, List<String> argTaskIds) {
+    if (LOGGER.isDebugEnabled()) {
+      LOGGER.debug(
+          "entry to setOwnerOfTasks(owner = {}, tasks = {})",
+          owner,
+          LoggerUtils.listToString(argTaskIds));
+    }
+    try {
+      taskanaEngine.openConnection();
```
holgerhagen: We could move this down to do the preparations before we open the connection. This way we reduce the time the DB connection is used.

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -617,6 +619,55 @@ public void forceDeleteTask(String taskId)
...
+      BulkOperationResults<String, TaskanaException> bulkLog = new BulkOperationResults<>();
+      if (argTaskIds == null || argTaskIds.isEmpty()) {
+        return bulkLog;
+      }
+      // remove duplicates
+      List<String> taskIds = argTaskIds.stream().sorted().distinct().collect(Collectors.toList());
```
holgerhagen: why sorted?

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -617,6 +619,55 @@ public void forceDeleteTask(String taskId)
...
+      final int requestSize = taskIds.size();
+      // use only elements we are authorized for
+      taskIds = filterForAuthorized(taskIds, bulkLog);
+      // set the Owner of these tasks we are authorized for
+      if (taskIds.isEmpty()) {
+        return bulkLog;
+      } else {
+        final int numberOfAffectedTasks = taskMapper.setOwnerOfTasks(owner, taskIds, Instant.now());
+        // check the outcome
+        List<MinimalTaskSummary> existingMinimalTaskSummaries =
```
holgerhagen: Cannot we do the authorization check withon the retrieval of the tasks?

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -757,6 +808,77 @@ Duration calculateDuration(
...
+          bulkLog.addError(
+              taskSummary.getTaskId(),
+              new UpdateFailedException(
+                  "Could not set owner of Task " + taskSummary.getTaskId() + "."));
+        }
+      }
+    }
+  }
+
+  private void handleNonExistingTasks(
```
holgerhagen: Personally I think we could be more precise in the naming. "AddExceptionForNonExistingTasks" for example. And isn't it a good practice to return the modified value? In this case the bulklog?

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
BerndBreier: I had read that distinct() works only correctly on ordered streams. However, this is not true. 
I will remove the sorted() method.
The javadoc says:
Stream<T> distinct()

Returns a stream consisting of the distinct elements (according to Object.equals(Object)) of this stream.
For ordered streams, the selection of distinct elements is stable (for duplicated elements, the element appearing first in the encounter order is preserved.) For unordered streams, no stability guarantees are made.

This is a stateful intermediate operation.

API Note:
    Preserving stability for distinct() in parallel pipelines is relatively expensive (requires that the operation act as a full barrier, with substantial buffering overhead), and stability is often not needed. Using an unordered stream source (such as generate(Supplier)) or removing the ordering constraint with BaseStream.unordered() may result in significantly more efficient execution for distinct() in parallel pipelines, if the semantics of your situation permit. If consistency with encounter order is required, and you are experiencing poor performance or memory utilization with distinct() in parallel pipelines, switching to sequential execution with BaseStream.sequential() may improve performance.
Returns:
    the new stream

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
BerndBreier: We can, but then we cannot distinguish the reason why an update failed - due to non-existence or due to missing authorization.

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
BerndBreier: Ok, I checked - lowercasing is done in CurrentUserContext. I removed it from TaskServiceImpl

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
BerndBreier: Done

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
BerndBreier: Done. I changed all new methods so that they return the bulklog that they create locally.

BerndBreier:
SonarCloud is green:
https://sonarcloud.io/dashboard?branch=TSK-1131&id=BerndBreier_taskana

BerndBreier:
Finally, Sonarcloud is error-free
https://sonarcloud.io/dashboard?branch=TSK-1131&id=BerndBreier_taskana

PR #964, opened 2020-03-05
TSK-1149: fixed test so that the build will also run on macos

sfrevel:
- fixed the test case
- removed unused code
- added java environment configuration file for jenv in .gitignore

### DIFF
--- a/lib/taskana-core/src/test/java/acceptance/task/SetPlannedAccTest.java
+++ b/lib/taskana-core/src/test/java/acceptance/task/SetPlannedAccTest.java
@@ -14,13 +14,13 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 
-import pro.taskana.classification.api.ClassificationService;
 import pro.taskana.classification.api.exceptions.ClassificationNotFoundException;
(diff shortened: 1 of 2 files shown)


lib/taskana-core/src/test/java/acceptance/task/SetPlannedAccTest.java
```diff
@@ -34,12 +34,11 @@
 @SuppressWarnings({"checkstyle:LineLength"})
 public class SetPlannedAccTest extends AbstractAccTest {
   private TaskService taskService;
-  private ClassificationService classificationService;
 
   @BeforeEach
   void setup() throws SQLException {
+    DaysToWorkingDaysConverter.setGermanPublicHolidaysEnabled(true);
```
mustaphazorgati: Stupid question. Couldn't we just explicitly disable the holidays?
Honestly I don't know if we want to have the Converter running in the background. If there is an error in the converter more tests would not be able to finish successfully

sfrevel: There are other test in the package pro.taskana.monitor.internal which are also enabling the german holidays. As far as i understood these tests test the german holidays explicitly and because the setter for enabling is static it is set for all test which are running afterwards also. 
I guess disabling the the german holidays explicitly before this test in SetPlannedAccTest  should also be an solution.

PR #973, opened 2020-03-12
Tsk 442

BerndBreier:


lib/taskana-core/src/main/java/pro/taskana/TaskanaEngineConfiguration.java
```diff
@@ -135,6 +135,8 @@ public TaskanaEngineConfiguration(
       this.rolesSeparator = rolesSeparator;
     }
 
+    setGermanPublicHolidaysEnabled(true);
```
holgerhagen: This should obviously be configurable, shouldn't it? Otherwise the switch does not make sense.

lib/taskana-core/src/main/java/pro/taskana/common/internal/jobs/JobRunner.java
```diff
@@ -114,6 +149,24 @@ private void runScheduledJob(ScheduledJob scheduledJob) {
               + e.getMessage(),
           e);
     }
-    LOGGER.debug("exit from runScheduledJob");
+  }
+
+  private Subject getAdminSubject() {
```
holgerhagen: We do have TaskanaEngineImpl.runAsAdmin(). Do you think we could reuse this here instead of creating a separate method?

lib/taskana-core/src/main/java/pro/taskana/TaskanaEngineConfiguration.java
BerndBreier: The setGermanPublicHolidaysEnabled() method is public.It is the interface, the customer can use to configure the TaskanaEngineConfiguration with respect to german holidays.

The referenced call simply sets the default to true. Without this call, the default is false.
Since the product is developed and mostly used in Germany, I think that **true** is the correct default...

lib/taskana-core/src/main/java/pro/taskana/common/internal/jobs/JobRunner.java
BerndBreier: This was my first idea: use InternalTaskanaEngineImpl.runAsAdmin. However, that method is a method on InternalTaskanaEngineImpl and JobRunner  has no access to InternalTaskanaEngineImpl, only to TaskanaEngineImpl.

PR #988, opened 2020-03-24
Tsk 1160

BerndBreier:


lib/taskana-core/src/main/java/pro/taskana/task/api/TaskService.java
```diff
@@ -490,4 +490,28 @@ TaskComment getTaskComment(String taskCommentId)
    */
   BulkOperationResults<String, TaskanaException> setPlannedPropertyOfTasks(
       Instant planned, List<String> taskIds);
+
+  /**
+   * Cancels a task.
```
holgerhagen: Cancels a task.
Cancellation means a task is obsolete from a business perspective an does not need to be completed anymore.

lib/taskana-core/src/main/java/pro/taskana/task/api/TaskService.java
```diff
@@ -490,4 +490,28 @@ TaskComment getTaskComment(String taskCommentId)
...
+   * @return the updated task.
+   * @throws TaskNotFoundException if the Task with Id TaskId is not found
+   * @throws InvalidStateException if the task is not in state READY or CLAIMED
+   * @throws NotAuthorizedException if the current user is not authorized to see the task
+   */
+  Task cancelTask(String taskId)
+      throws TaskNotFoundException, InvalidStateException, NotAuthorizedException;
+
+  /**
+   * Terminates a task.
```
holgerhagen: Terminates a task.
Termination is a administrative action to complete a task. This is typically done by an administration to correct any technical issue.

lib/taskana-core/src/main/java/pro/taskana/task/api/TaskState.java
```diff
@@ -4,5 +4,20 @@
 public enum TaskState {
   READY,
   CLAIMED,
-  COMPLETED
+  COMPLETED,
+  CANCELLED,
+  TERMINATED;
+
+  public boolean isInStates(TaskState... states) {
```
holgerhagen: Maybe just "in" as it is a well known operator for this? Just an idea.

lib/taskana-core/src/main/java/pro/taskana/task/internal/ServiceLevelHandler.java
```diff
@@ -48,14 +46,9 @@
...
-    DaysToWorkingDaysConverter.setGermanPublicHolidaysEnabled(
-        taskanaEngine.getEngine().getConfiguration().isGermanPublicHolidaysEnabled());
-    try {
-      converter = DaysToWorkingDaysConverter.initialize();
-    } catch (InvalidArgumentException e) {
-      LOGGER.error(ERROR_CANNOT_INITIALIZE_DAYS_TO_WORKING_DAYS_CONVERTER);
-      throw new SystemException(
-          ERROR_CANNOT_INITIALIZE_DAYS_TO_WORKING_DAYS_CONVERTER, e.getCause());
+    this.converter = WorkingDaysToDaysConverter.initialize();
+    if (taskanaEngine.getEngine().getConfiguration().isGermanPublicHolidaysEnabled()) {
```
holgerhagen: Why don't we provide the converter on engine level and initialize it during startup according to the configuration?

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -951,6 +979,29 @@ private TaskImpl checkConcurrencyAndSetModified(TaskImpl newTaskImpl, TaskImpl o
...
+    TaskState state = task.getState();
+    if (state.isEndState()) {
+      throw new InvalidStateException(String.format(TASK_WITH_ID_IS_ALREADY_IN_END_STATE, taskId));
+    }
+
+    Instant now = Instant.now();
+    task.setOwner(null);
+    task.setModified(now);
+    task.setCompleted(now);
+    task.setClaimed(null);
```
holgerhagen: We should retain this timestamp even if the task is terminated/cancelled.

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -951,6 +979,29 @@ private TaskImpl checkConcurrencyAndSetModified(TaskImpl newTaskImpl, TaskImpl o
...
+    task.setRead(true);
```
holgerhagen: Why this. We don't know if the task has been read, do we?

lib/taskana-core/src/test/java/acceptance/task/CancelTaskAccTest.java
```diff
@@ -0,0 +1,128 @@
...
+      throws NotAuthorizedException, TaskNotFoundException, InvalidStateException {
+    List<TaskSummary> taskSummaries =
+        taskService.createTaskQuery().stateIn(TaskState.CLAIMED).list();
+    assertThat(taskSummaries.size()).isEqualTo(19);
+    Task task = taskService.getTask(taskSummaries.get(0).getId());
+    taskService.cancelTask(taskSummaries.get(0).getId());
+    long numTasks = taskService.createTaskQuery().stateIn(TaskState.CLAIMED).count();
+    assertThat(numTasks).isEqualTo(18);
+    numTasks = taskService.createTaskQuery().stateIn(TaskState.CANCELLED).count();
+    assertThat(numTasks).isEqualTo(6);
```
holgerhagen: Maybe we should add an assertion isEqualTo(5) before cancelTask to make it more clear?

lib/taskana-core/src/test/java/acceptance/task/CancelTaskAccTest.java
```diff
@@ -0,0 +1,128 @@
...
+      groupNames = {"group_1"})
+  @Test
+  void testQeryCancelledTasks() {
+    List<TaskSummary> taskSummaries =
+        taskService.createTaskQuery().stateIn(TaskState.CANCELLED).list();
+    assertThat(taskSummaries.size()).isEqualTo(5);
+  }
+
+  @WithAccessId(
+      userName = "admin",
```
holgerhagen: Is it a good idea to do all these tests as admin?

lib/taskana-core/src/test/java/acceptance/task/CancelTaskAccTest.java
```diff
@@ -0,0 +1,128 @@
...
+    Task task = taskService.getTask(taskSummaries.get(0).getId());
+    taskService.cancelTask(taskSummaries.get(0).getId());
+    long numTasks = taskService.createTaskQuery().stateIn(TaskState.READY).count();
+    assertThat(numTasks).isEqualTo(46);
+    numTasks = taskService.createTaskQuery().stateIn(TaskState.CANCELLED).count();
+    assertThat(numTasks).isEqualTo(6);
+  }
+
+  @WithAccessId(
+      userName = "admin",
```
holgerhagen: Did we decide that everybody is allowed to cancel a claimed task? I wonder if it would be better to allow only the owner to cancel claimed tasks.

lib/taskana-core/src/test/java/acceptance/task/TerminateTaskAccTest.java
```diff
@@ -0,0 +1,111 @@
...
+import pro.taskana.task.api.TaskService;
+import pro.taskana.task.api.TaskState;
+import pro.taskana.task.api.exceptions.InvalidStateException;
+import pro.taskana.task.api.exceptions.TaskNotFoundException;
+import pro.taskana.task.api.models.Task;
+import pro.taskana.task.api.models.TaskSummary;
+
+/** Acceptance tests for all claim and complete scenarios. */
+@ExtendWith(JaasExtension.class)
+class TerminateTaskAccTest extends AbstractAccTest {
```
holgerhagen: See comments on cancel task test.

lib/taskana-core/src/main/java/pro/taskana/task/api/TaskService.java
BerndBreier: I added the description

lib/taskana-core/src/main/java/pro/taskana/task/api/TaskService.java
BerndBreier: I added the description

lib/taskana-core/src/main/java/pro/taskana/task/internal/ServiceLevelHandler.java
BerndBreier: Apart from the DaysToWorkingDaysReportConverter, the converter is only needed in ServiceLevelHandler. I think, we should maintain and supply classes / objects there where they are needed. But if you insist to provide it on engine level, I can do that.
With respect to configuration: I introduce the configuration property taskana.german.holidays.enabled and set it to true in taskana.properties. Then I initialize this property in TaskanaEngineConfiguration and use it in the constructor of TaskanaEngineImpl to set that (static) property on WorkingDaysToDaysConverter.

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
BerndBreier: ok, retained the claimed timestamp

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
BerndBreier: ok, retained also the read timestamp

lib/taskana-core/src/test/java/acceptance/task/CancelTaskAccTest.java
BerndBreier: Done

lib/taskana-core/src/test/java/acceptance/task/CancelTaskAccTest.java
BerndBreier: Changed it ...

lib/taskana-core/src/test/java/acceptance/task/CancelTaskAccTest.java
BerndBreier: The ticket says : " terminateTask() and cancelTask bay be called by every user that is allowed to see the task " 
I agree, that it is strange that everybody is allowed to cancel / terminate a task. On the other hand, if you think about the adapter scenario: if a task has been cancelled / terminated in camunda, the adapter should be able to cancel / terminate it in taskana. And this without being the owner. And probably also without having admin authority. So before I change anything with respect to this, we should carefully think about what is the correct solution ....

lib/taskana-core/src/test/java/acceptance/task/TerminateTaskAccTest.java
BerndBreier: done

lib/taskana-core/src/main/java/pro/taskana/task/internal/ServiceLevelHandler.java
```diff
@@ -48,15 +46,7 @@
...
-    DaysToWorkingDaysConverter.setGermanPublicHolidaysEnabled(
-        taskanaEngine.getEngine().getConfiguration().isGermanPublicHolidaysEnabled());
-    try {
-      converter = DaysToWorkingDaysConverter.initialize();
-    } catch (InvalidArgumentException e) {
-      LOGGER.error(ERROR_CANNOT_INITIALIZE_DAYS_TO_WORKING_DAYS_CONVERTER);
-      throw new SystemException(
-          ERROR_CANNOT_INITIALIZE_DAYS_TO_WORKING_DAYS_CONVERTER, e.getCause());
-    }
+    this.converter = WorkingDaysToDaysConverter.initialize();
```
holgerhagen: Why don't we move this to the engine as well?

lib/taskana-core/src/main/java/pro/taskana/task/api/TaskState.java
holgerhagen: You don't like that? Or you has overseen this ons?

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -951,6 +979,27 @@ private TaskImpl checkConcurrencyAndSetModified(TaskImpl newTaskImpl, TaskImpl o
...
+      throw new TaskNotFoundException(taskId, String.format(TASK_WITH_ID_WAS_NOT_FOUND, taskId));
+    }
+    TaskImpl task = (TaskImpl) getTask(taskId);
+    TaskState state = task.getState();
+    if (state.isEndState()) {
+      throw new InvalidStateException(String.format(TASK_WITH_ID_IS_ALREADY_IN_END_STATE, taskId));
+    }
+
+    Instant now = Instant.now();
+    task.setOwner(null);
```
holgerhagen: Sorry, haven't seen this one right away. As we don't reset the claimed timestamp, we shouldn't reset the owner as well.

rest/taskana-rest-spring-example-common/src/test/resources/taskana.properties
```diff
@@ -12,3 +12,4 @@ taskana.jobs.batchSize=50
 taskana.jobs.cleanup.runEvery=P1D
 taskana.jobs.cleanup.firstRunAt=2018-07-25T08:00:00Z
 taskana.jobs.cleanup.minimumAge=P14D
+taskana.german.holidays.enabled=true
```
holgerhagen: But didn't you changed this to be the default? Why do we have to add it to all properties now?

lib/taskana-core/src/main/java/pro/taskana/task/api/TaskState.java
BerndBreier: Ok, I didn't see this one. No problem, I'll change it to 'in'

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
BerndBreier: Done

lib/taskana-core/src/main/java/pro/taskana/task/internal/ServiceLevelHandler.java
BerndBreier: The engine calls only WorkingDaysToDaysConverter.setGermanPublicHolidaysEnabled() which is a static method. The initialize() call is a factory method that creates an instance of WorkingDaysToDaysConverter.
Since the WorkingDaysToDaysConverter is only needed inside ServiceLevelHandler, I made it an instance variable of this class.
To make things consistent, I'll also move the setGermanPublicHolidaysEnabled() call to the ctor of ServiceLevelHandler.

rest/taskana-rest-spring-example-common/src/test/resources/taskana.properties
BerndBreier: You requested to make this configurable. Therefore, I introduced configuration property taskana.german.holidays.enabled, which is a boolean. Usually the default value for a boolean property is false. Therefore, I didn't want to set this property to true if it is not specified. But if you say, I should, then I'll do this.

PR #1004, opened 2020-04-03
TSK-746 Add simple history event tracking to update and create task m…

gitgoodjhe:
…ethod


gitgoodjhe:
SonarCloud does NOT pass due to test coverage. This is incorrect though, since it's the same problem the TASKANA-Adapter had: A seperate testing module. So the test results don't get aggregated yet. TSK-1035 should fix thix, so it doesn't make sense to use the solution from the adapter. 

https://sonarcloud.io/dashboard?branch=TSK-746&id=gitgoodjhe_taskana

The security hotspot is raised by default when using reflection and says "Make sure that this accessibility update is safe here." - which it is!

history/taskana-simplehistory-provider/src/test/java/acceptance/query/QueryHistoryAccTest.java
```diff
@@ -25,19 +25,19 @@ public QueryHistoryAccTest() {
   @Test
   public void testListValuesAscendingAndDescending() {
     List<String> defaultList =
-        getHistoryService().createHistoryQuery().listValues(HistoryQueryColumnName.COMMENT, null);
+        getHistoryService().createHistoryQuery().listValues(HistoryQueryColumnName.TASK_ID, null);
```
holgerhagen, line 28: Since this is an acceptance test, I would use a more useful testcase like sorting by 'created'.

history/taskana-simplehistory-provider/src/main/java/pro/taskana/simplehistory/query/HistoryQueryColumnName.java
```diff
@@ -26,16 +26,13 @@
...
   OLD_VALUE("old_value"),
   NEW_VALUE("new_value"),
   CUSTOM_1("custom_1"),
   CUSTOM_2("custom_2"),
   CUSTOM_3("custom_3"),
   CUSTOM_4("custom_4"),
-  OLD_DATA("old_data"),
-  NEW_DATA("new_data"),
-  TYPE("type");
+  TYPE("event_Type");
```
holgerhagen, line 35: 'event_type'?

history/taskana-simplehistory-provider/src/test/java/acceptance/query/QueryHistoryAccTest.java
```diff
@@ -25,19 +25,19 @@ public QueryHistoryAccTest() {
   @Test
   public void testListValuesAscendingAndDescending() {
```
holgerhagen, line 26: could you change the name of the tests to the new schema, please?

history/taskana-simplehistory-provider/src/test/java/pro/taskana/simplehistory/impl/HistoryQueryImplTest.java
```diff
@@ -66,13 +65,12 @@ private HistoryEventImpl createHistoryEvent(
       String workbasketKey,
       String type,
       String userId,
-      String comment,
+      String details,
       Instant created) {
-    HistoryEventImpl he = new HistoryEventImpl(userId);
+    HistoryEventImpl he = new HistoryEventImpl(userId, details);
```
holgerhagen, line 70: what happens to the details?

lib/taskana-core/src/main/java/pro/taskana/spi/history/api/events/task/ClaimCancelledEvent.java
```diff
@@ -6,7 +6,7 @@
 public class ClaimCancelledEvent extends TaskEvent {
 
   public ClaimCancelledEvent(Task task, String userId) {
-    super(task, userId);
+    super(task, userId,"");
```
holgerhagen, line 9: should we rather use 'null'? I just don't want that the database created an empty CLOB for an empty String. Do you know how it is handled?

lib/taskana-core/src/main/java/pro/taskana/spi/history/api/TaskanaHistory.java
```diff
@@ -20,4 +21,14 @@
...
   void create(TaskanaHistoryEvent event);
+
+  /**
+   * Get the details of a history event by Id.
+   *
+   * @param historyEventId the id of the history event
+   * @return the history event
+   * @throws TaskanaHistoryEventNotFoundException If the Id rfers to a not existing history event
+   */
+  TaskanaHistoryEvent getHistoryEvent(String historyEventId)
```
holgerhagen, line 32: I don't think that this should be part of the SPI. This would create a dependency from core to the history we don't want to have.

lib/taskana-data/src/main/resources/sql/sample-data/history-event.sql
```diff
@@ -1,52 +1,47 @@
...
+-- BUSINESS_PROCESS_ID, PARENT_BUSINESS_PROCESS_ID, TASK_ID,                                        EVENT_TYPE, CREATED,                                        USER_ID,    DOMAIN,         WORKBASKET_KEY,                                                     POR_COMPANY     , POR_SYSTEM, POR_INSTANCE      , POR_TYPE      , POR_VALUE     , TASK_CLASSIFICATION_KEY, TASK_CLASSIFICATION_CATEGORY , ATTACHMENT_CLASSIFICATION_KEY , OLD_VALUE     , NEW_VALUE     , CUSTOM_1      , CUSTOM_2      , CUSTOM_3      , CUSTOM_4
+('BPI:01'               ,'',                        'TKI:000000000000000000000000000000000000',     'CREATE',   CURRENT_TIMESTAMP ,                      'USER_2_2', 'DOMAIN_B',         'WBI:100000000000000000000000000000000001', '00'                , 'PASystem', '00'                      , 'VNR'         , '11223344'                     ,' L140101'                         , 'TASK'                       ,''                    ,'old_val12'      ,'new_val12'      ,'custom1'      ,'custom2'      , 'custom3'     ,'custom4', 'some Details'),
```
holgerhagen, line 3: I'm not sure but haven't you changed the type to 'TASK_CREATED' instead of 'CREATE'? Did I get this wrong?

lib/taskana-core/src/main/java/pro/taskana/spi/history/api/TaskanaHistory.java
gitgoodjhe: Do you have a suggestion where I should put this alternatively?

history/taskana-simplehistory-provider/src/test/java/acceptance/query/QueryHistoryAccTest.java
gitgoodjhe: Done

history/taskana-simplehistory-provider/src/test/java/pro/taskana/simplehistory/impl/HistoryQueryImplTest.java
gitgoodjhe: I'm not sure what you mean. HistoryEventImpl extends TaskanaHistoryEvent which has a constructor that sets the details (as well as the userId).

lib/taskana-core/src/main/java/pro/taskana/spi/history/api/events/task/ClaimCancelledEvent.java
gitgoodjhe: Was considering null too, changed this now

lib/taskana-data/src/main/resources/sql/sample-data/history-event.sql
gitgoodjhe: Yes, this is old Test-data. I adjusted it now

history/taskana-simplehistory-provider/src/main/java/pro/taskana/simplehistory/query/HistoryQueryColumnName.java
gitgoodjhe: I removed this since there is no such field "type" in the database, only "event_type" but now I saw that "event_type" is already defined some lines before that

history/taskana-simplehistory-provider/src/test/java/acceptance/query/QueryHistoryAccTest.java
gitgoodjhe: Done

lib/taskana-data/src/main/java/pro/taskana/sampledata/SampleDataGenerator.java
```diff
@@ -69,11 +69,15 @@ public void generateSampleData() {
           }
           executeAndCacheScripts(scripts, cacheKey);
         });
-    LOGGER.debug("exit from generateSampleData()");
+    if (LOGGER.isDebugEnabled()) {
```
holgerhagen, line 72: We do not need the guarded logging here, as this is the first call in the logger.

lib/taskana-core/src/main/java/pro/taskana/common/internal/util/FieldAndValuePairTriplet.java
```diff
@@ -0,0 +1,40 @@
+package pro.taskana.common.internal.util;
+
+import java.lang.reflect.Field;
+
+public class FieldAndValuePairTriplet {
+
```
mustaphazorgati, line 6: Do you mind using our generic triplet class?

gitgoodjhe: Sure, changed it

PR #1010, opened 2020-04-10
TSK-1182: refined architecture tests

mustaphazorgati:


mustaphazorgati:
Hello @holgerhagen!
That's true. We now have those dependencies. Unfortunately there is no way around it since the class `AbstractTaskanaJob` is a factory and produced the specific Jobs. For that we need to have a dependency to all of its subclasses. Personally I think this is ok since the design pattern enforces us to have that dependency. 

Now the `common` package has only two non-common classes: TaskanaEngine and AbstractTaskanaJob. Every other class has no dependency outside of common and can be verified.

An alternative we thought about is to exclude the entire jobs package from the architecture test. Again, we chose out of personal preference.

mustaphazorgati:
@gitgoodjhe what do you think about this change? Is it as critical as holger describes?

gitgoodjhe:
This really comes down to taste I think. Personally I would have excluded the jobs package from the architecture test.

mustaphazorgati:
> This really comes down to taste I think. Personally I would have excluded the jobs package from the architecture test.

Allright. will do. @holgerhagen is that fine for you too?

lib/taskana-core/src/main/java/pro/taskana/task/internal/jobs/TaskRefreshJob.java
```diff
@@ -1,15 +1,17 @@
-package pro.taskana.common.internal.jobs;
+package pro.taskana.task.internal.jobs;
 
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import pro.taskana.classification.internal.jobs.ClassificationChangedJob;
```
holgerhagen, line 9: This dependency is only required because of some constants. Do we really want that? In this case I would get reid of the dependency and declare the constants again. What do you think?

mustaphazorgati: Done.

Now we only have dependencies from common.internal.jobs -> [task|workbasket|classification].internal.jobs 

This is necessary, since the AbstractTaskanaJob is a JobFactory

PR #1012, opened 2020-04-13
TSK-1183: Update to Spring Boot 2.2.6

holgerhagen:


lib/taskana-data/src/test/java/pro/taskana/sampledata/SampleDataProviderTest.java
```diff
@@ -10,19 +10,19 @@
...
   void getScriptsNotNull() {
-    assertThat(SampleDataProvider.getDefaultScripts()).isNotNull();
+    assertThat(SampleDataProvider.getSampleDataCreationScripts()).isNotNull();
     assertThat(SampleDataProvider.getScriptsWithEvents()).isNotNull();
   }
 
   @Test
   void getScriptsNotEmpty() {
-    assertThat(SampleDataProvider.getDefaultScripts().count() > 0).isTrue();
+    assertThat(SampleDataProvider.getSampleDataCreationScripts().count() > 0).isTrue();
```
mustaphazorgati, line 19: Please replace with
assertThat(SampleDataProvider.getSampleDataCreationScripts().count()).isGreaterThan(0);

-- Can you please also do this for the line below? And maybe if it is somewhere else aswell? That'd be great

holgerhagen: Actually I haven't changed anything here. I just renamed the method. If you would like to change something, please open a ticket for it.

holgerhagen: Regarding the logging: this is perfect the way it is as long as we do not create any Strings in the method call parameters.

PR #1013, opened 2020-04-14
TSK-1188 Enable combo filter for tasks

gitgoodjhe:
sonarcloud is green: 

https://sonarcloud.io/dashboard?branch=TSK-1188&id=gitgoodjhe_taskana


rest/taskana-rest-spring/src/main/java/pro/taskana/rest/TaskController.java
```diff
@@ -442,6 +468,22 @@ private void checkForIllegalParamCombinations(MultiValueMap<String, String> para
               + PLANNED_UNTIL
               + "\"");
     }
+
+    if (params.containsKey(WILDCARD_SEARCH_FIELDS) && !params.containsKey(WILDCARD_SEARCH_VALUE)
+        || !params.containsKey(WILDCARD_SEARCH_FIELDS)
+            && params.containsKey(WILDCARD_SEARCH_VALUE)) {
+
+      throw new IllegalArgumentException(
+          "It is prohibited to use the params "
```
holgerhagen, line 477: I think the message would be more understandable if you only point out that both parameters have to be present if used.

lib/taskana-core/src/main/java/pro/taskana/task/api/TaskQuery.java
```diff
@@ -611,6 +611,23 @@ TaskQuery customAttributeLike(String num, String... searchArguments)
...
+   * Add your wildcard search value for pattern matching to your query. It will be compared in SQL
+   * with the LIKE operator. You may use a wildcard like % to specify the pattern.
+   *
+   * @param wildcardSearchValue the wildcard search value
+   * @return the query
+   */
+  TaskQuery wildcardSearchValueLike(String wildcardSearchValue);
+
+  /**
+   * Add the values of the wildcard search fields for exact matching to your query.
```
holgerhagen, line 624: Maybe we should add a hint that both parameters need to be applied in parallel?

PR #1041, opened 2020-04-29
TSK-1205, TSK-1201 Replaced the old alerts, errors and other popups with a central solution

Tristan2357:
The old Alerts had a bigger overhead, which is why they got replaced by a toast. This Toast removed some clutter both from the main app.component as from the entire structure of the project.
Previously we had three components for displaying a popup to the user. This has been reduced to one dialog-pop-up component which handles all of the uses. The one component, which the new popup does not replace is the GeneralModalComponent. This was only used in the history and served little to no purpose there. 
 Top Left: The old PopUp; Top Right: The old interface (the blue icons open the popup); Bottom: The new version. It just shows the corresponding values in the table instead of opening a popup


web/package.json
```diff
@@ -30,6 +32,7 @@
     "chart.js": "2.7.2",
     "core-js": "2.5.3",
     "file-saver": "1.3.3",
+    "hammerjs": "^2.0.8",
```
cnguyen-de, line 35: please remove hammerjs from package json since it's only for controlling touch behaviors

cnguyen-de:
should be fine for merging after removing hammerjs in package.json

PR #1046, opened 2020-05-03
fixed newly introduced code smells

mustaphazorgati:
fixed the 3 new introduced code smells https://sonarcloud.io/project/issues?id=Taskana_taskana&resolved=false&sinceLeakPeriod=true&types=CODE_SMELL


lib/taskana-core/src/main/java/pro/taskana/monitor/api/reports/header/TimeIntervalColumnHeader.java
```diff
@@ -34,19 +34,15 @@ public TimeIntervalColumnHeader(int lowerAgeLimit, int upperAgeLimit) {
   public static int getSmallestUpperLimit(List<? extends TimeIntervalColumnHeader> columnHeaders) {
     return columnHeaders.stream()
         .mapToInt(TimeIntervalColumnHeader::getUpperAgeLimit)
-        .filter(i -> i < 0)
```
sfrevel, line 37: I guess the filter is useless?

PR #1064, opened 2020-05-13
TSK-1218: Prevent unsecure access to TASKANA

gitgoodjhe:
Sonarcloud: https://sonarcloud.io/dashboard?branch=TSK-1218&id=gitgoodjhe_taskana


lib/taskana-core/src/main/java/pro/taskana/common/internal/configuration/SecurityVerifier.java
```diff
@@ -0,0 +1,97 @@
...
+
+  public SecurityVerifier(DataSource dataSource, String schema) {
+    super();
+    this.dataSource = dataSource;
+    this.schemaName = schema;
+  }
+
+  public void checkSecureAccess(boolean securityEnabled) {
+
+    if (!securityEnabled) {
```
holgerhagen, line 31: why do you check this? Even if security is activated we might be the first one and want to insert this information into the database, don't we?

lib/taskana-core/src/test/java/acceptance/config/TaskanaSecurityConfigAccTest.java
```diff
@@ -0,0 +1,134 @@
...
+          TaskanaEngineTestConfiguration.getSchemaName());
+    };
+
+    assertThatCode(createUnsecuredTaskanaEngineConfiguration).doesNotThrowAnyException();
+
+    assertThat(retrieveSecurityFlag()).isFalse();
+
+  }
+
+  private Boolean retrieveSecurityFlag() throws SQLException {
```
holgerhagen, line 96: According to my comment above: do we miss the test for starting secure the first time and initialize the security flag?

PR #1065, opened 2020-05-13
TSK-1246: Fix updateTaskComment in TaskCommentService

gitgoodjhe:
Sonarcloud: https://sonarcloud.io/dashboard?branch=TSK-1246&id=gitgoodjhe_taskana

### DIFF
--- a/lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskCommentServiceImpl.java
+++ b/lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskCommentServiceImpl.java
@@ -69,13 +69,13 @@ class TaskCommentServiceImpl {
 
       taskService.getTask(taskCommentImplToUpdate.getTaskId());
 
-      if (taskCommentToUpdate.getCreator().equals(userId)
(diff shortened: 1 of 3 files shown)


lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskCommentServiceImpl.java
```diff
@@ -69,13 +69,13 @@ TaskComment updateTaskComment(TaskComment taskCommentToUpdate)
 
       taskService.getTask(taskCommentImplToUpdate.getTaskId());
 
-      if (taskCommentToUpdate.getCreator().equals(userId)
+      TaskComment originalTaskComment = getTaskComment(taskCommentImplToUpdate.getId());
```
holgerhagen, line 72: We already do a getTask() in getTaskComment() (I guess in order to check whether the user is authorized to see the task?) In this case the getTask() in line 70 is unnecessary, isn't it?

gitgoodjhe: That's true. Removed the additional getTask call from line 70

PR #1069, opened 2020-05-13
TSK-1205 Added color to Toasts and fixed a poblem where multiple alerts were shown simultaneously

Tristan2357:
The Toasts now have a color depending on the type of alert (e.g. success -> green). There were cases, were errors led to multiple Alerts simultaneously. This was fixed (hopefully :)).


web/src/app/shared/components/popup/dialog-pop-up.component.ts
```diff
@@ -32,8 +32,8 @@ export class DialogPopUpComponent implements OnInit {
   }
 
   initError() {
-    this.title = notifications.get(this.data.key).name;
-    this.message = notifications.get(this.data.key).text || this.data.passedError.error.message;
+    this.title = notifications.get(this.data.key).name || '';
+    this.message = notifications.get(this.data.key).text || this.data.passedError.error ? this.data.passedError.error.message : '';
```
mustaphazorgati, line 36: Why did you introduce the ternary operator?

web/src/app/shared/services/http-client-interceptor/http-client-interceptor.service.ts
```diff
@@ -29,7 +29,7 @@ export class HttpClientInterceptor implements HttpInterceptor {
       } else if (error instanceof HttpErrorResponse && (error.status === 404) && error.url.indexOf('environment-information.json')) {
         // ignore this error
       } else {
-        this.errorsService.triggerError(NOTIFICATION_TYPES.GENERAL_ERR, error);
+        // this.errorsService.triggerError(NOTIFICATION_TYPES.GENERAL_ERR, error);
```
mustaphazorgati, line 32: Do we really want to ignore all errors from the interceptor?

web/src/app/shared/components/popup/dialog-pop-up.component.ts
Tristan2357: Because sometimes passedError is undefined and will cause an error, when we check passedError.error.message  just via an OR. 
These errors were not really important, but they were annoyingly filling up the console

web/src/app/shared/services/http-client-interceptor/http-client-interceptor.service.ts
Tristan2357: This could be confusing for users, but nearly all expectable errors should be covered on a more specific level.
The alternative would be to either check if there already is an alert (which is difficult because this general error is always covered before the specific one) or we check if the received error is one that we expect elsewhere (i personally could not think of an elegant and scalable way to do this) or last option which would be huge rework: put all http error handling in this function and show the coresponding error for every case (lots of work without real benefit or an elegant abd scalable solution)

PR #1097, opened 2020-06-03
TSK-1247: Removed LoggerUtils and replaced its usages with the standard toString implementations

Tristan2357:
This has been dicussed in the [Community of Pratices](https://taskana.atlassian.net/wiki/spaces/TAS/pages/883228673/Community+of+practices)


lib/taskana-core/src/main/java/pro/taskana/common/internal/TaskanaEngineImpl.java
```diff
@@ -209,7 +208,7 @@ public boolean isUserInRole(TaskanaRole... roles) {
   public void checkRoleMembership(TaskanaRole... roles) throws NotAuthorizedException {
     if (!isUserInRole(roles)) {
       if (LOGGER.isDebugEnabled()) {
-        String accessIds = LoggerUtils.listToString(CurrentUserContext.getAccessIds());
+        String accessIds = CurrentUserContext.getAccessIds().toString();
```
mustaphazorgati, line 211: just use `CurrentUser.getAccessIds()` in the debug message. That way you don't have to use a `toString` call :)

Tristan2357: done :)

PR #1126, opened 2020-06-15
TSK-1039: Autofocus AccessID field after adding a new AccessID in Workbasket

Tristan2357:
> When administrating taskana accessIDs for a workbasket:
> 
> Given I want to create a new accessID
> When I click on the `+ Add new access` button
> Then I expect an empty accesItem entry (row) to be created AND I expect that the cursor is within the accessID so that I can immediately start typing the accessId after clicking the button.

This expected behaviour is now implemented.

### DIFF
--- a/web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.ts
+++ b/web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.ts
@@ -1,14 +1,21 @@
-import { Component, Input, OnChanges, OnDestroy, SimpleChanges } from '@angular/core';
+import { Component,
+  ElementRef,
+  Input,
+  OnChanges,
+  OnDestroy, QueryList,
+  SimpleChanges,
+  ViewChildren } from '@angular/core';
 import { Observable, Subscription } from 'rxjs';
 import { Select } from '@ngxs/store';
-import { FormArray, FormBuilder, Validators } from '@angular/forms';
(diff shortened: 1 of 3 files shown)


web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.ts
```diff
@@ -71,13 +84,10 @@ export class WorkbasketAccessItemsComponent implements OnChanges, OnDestroy {
...
   }
 
-  setAccessItemsGroups(accessItems: Array<WorkbasketAccessItems>) {
-    const AccessItemsFormGroups = accessItems.map(accessItem => this.formBuilder.group(accessItem));
-    AccessItemsFormGroups.forEach(accessItemGroup => {
-      accessItemGroup.controls.accessId.setValidators(Validators.required);
+  ngAfterViewInit() {
+    this.inputs.changes.subscribe(next => {
+      if (this.added) next.last.nativeElement.focus();
     });
```
mustaphazorgati, line 90: will this somehow collide with the function `focusNewInput`?

Tristan2357: i tested, if this could perhaps replace it, but it does not interact with it, as the input, that is used in focusNewInput is not in the domain of the workbasketAccessItems Component

PR #1138, opened 2020-06-22
TSK-1221: Added copy functionality to classifications

Tristan2357:
Added a Button to copy classifications.

### DIFF
--- a/web/src/app/administration/components/classification-details/classification-details.component.ts
+++ b/web/src/app/administration/components/classification-details/classification-details.component.ts
@@ -128,7 +128,7 @@ export class ClassificationDetailsComponent implements OnInit, OnDestroy {
     } else {
       this.store.dispatch(
         new RestoreSelectedClassification(this.classification.classificationId)
-      ).pipe(take(1)).subscribe(ctx => {
+      ).pipe(take(1)).subscribe(() => {
         this.notificationsService.showToast(NOTIFICATION_TYPES.INFO_ALERT);
       });
     }
@@ -136,7 +136,7 @@ export class ClassificationDetailsComponent implements OnInit, OnDestroy {
 
   onCopy() {
     this.store.dispatch(new SetActiveAction(ACTION.COPY));
-    this.classification.key = '';
+    this.classification.key = null;
   }
 
   selectCategory(category: string) {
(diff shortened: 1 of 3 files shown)


web/src/app/administration/components/classification-details/classification-details.component.ts
```diff
@@ -157,7 +165,8 @@ export class ClassificationDetailsComponent implements OnInit, OnDestroy {
 
   private async onSave() {
     this.requestInProgressService.setRequestInProgress(true);
-    if (this.action === ACTION.CREATE) {
+    if (this.action) {
+      this.classification.classificationId = null; // in case the id has been set, but a new classification should be created
```
mustaphazorgati, line 169: should this be done during save or creation? 
Personally I think that onCopy() should remove the Id.

web/src/app/administration/components/classification-details/classification-details.component.ts
```diff
@@ -131,6 +134,11 @@ export class ClassificationDetailsComponent implements OnInit, OnDestroy {
     }
   }
 
+  onCopy() {
+    this.store.dispatch(new SetActiveAction(ACTION.COPY));
+    this.classification.key = '';
```
mustaphazorgati, line 139: shouldn't we make a copy first? Are we otherwise overriding the values from the classification we copied from? Can you verify this please?

web/src/app/administration/components/classification-details/classification-details.component.ts
Tristan2357: This way the reset button can use the id to reset the copy to the original and in my opinion it does not really matter when to id gets removed. It would be cleaner, but i liked the possibility to reset.

web/src/app/administration/components/classification-details/classification-details.component.ts
Tristan2357: Because we use the store we always just work with a local copy, and when switching to the old classification we retrive the newest version of it, which resets unsaved changes. If saved the classification does not overwrite the old one

web/src/app/administration/components/classification-details/classification-details.component.ts
mustaphazorgati: Well.. the reset reset's the ID - I don't think this is desired..

PR #1140, opened 2020-06-22
TSK-1298: Enable drag'n drop of classifications in the tree

sofie29:
https://sonarcloud.io/dashboard?branch=TSK-1298&id=sofie29_taskana


mustaphazorgati:
@Tristan2357 and me looked over this PR and are not happy with the general way to handle this issue. Together with @cnguyen-de  we located the generation of the tree and moved it to a dedicated service. 
Our goal is to handle this issue on the client side and thus not create a further request to the rest service.

PR #1149, opened 2020-06-25
TSK-1306: Fixed problem with domain switching in classifications

Tristan2357:
https://sonarcloud.io/dashboard?branch=TSK-1306&id=Tristan2357_taskana

### DIFF
--- a/web/src/app/shared/services/domain/domain.service.ts
+++ b/web/src/app/shared/services/domain/domain.service.ts
@@ -64,9 +64,7 @@ export class DomainService {
   }
 
   switchDomain(value: string) {
-    this.requestInProgressService.setRequestInProgress(true);
     this.selectDomain(value);
-    this.router.navigate([this.getNavigationUrl()]);
   }
 
   /*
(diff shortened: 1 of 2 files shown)


web/src/app/shared/services/domain/domain.service.ts
```diff
@@ -64,9 +64,7 @@ export class DomainService {
   }
 
   switchDomain(value: string) {
-    this.requestInProgressService.setRequestInProgress(true);
     this.selectDomain(value);
-    this.router.navigate([this.getNavigationUrl()]);
```
mustaphazorgati, line 69: what's the reason for this change?

Tristan2357: Switching the domain should not affect the url, therefore the second call is not needed.
The selectDomain call is synchronous and should not need a spinner. The actual loading of the classifications or the workbaskets from another domain maybe should show a spinner, but in my opinion, this is not the responsibilitie of the domain service. If we were to start the spinner in this service, it should be understandable where it will stop it. this was not the case, and in most cases, the spinner was not stopped again, which caused the domainSwitching to fail in the first place.

Tristan2357: There would be solutions with the call for inProgress still there, would this be prefered? If yes this should be easily possible. 
Maybe i should start the inProgress at the inner calls? Or should i continue the current pattern?

mustaphazorgati: Let's talk about this. Maybe we need to refactor this service in order to get rid of this behaviour.

Currently this change leaves code fragments everywhere where `domainChangedComplete` is used.

Furthermore we have a behaviour change: previously changing the domain unselected the active classification / workbasket. This is not the case anymore

PR #1159, opened 2020-07-01
TSK-1253: Prevent LDAP injection.

holgerhagen:


rest/taskana-rest-spring/src/main/java/pro/taskana/common/rest/ldap/LdapClient.java
```diff
@@ -191,7 +191,18 @@ public AccessIdRepresentationModel searchGroupByDn(final String name) {
...
-    andFilter.and(new LikeFilter(getGroupsOfUser(), "*" + accessId + "*"));
+    final OrFilter orFilter = new OrFilter();
+    orFilter.or(new EqualsFilter(getGroupsOfUser(), accessId));
+    orFilter.or(
+        new EqualsFilter(
+            getGroupsOfUser(),
+            LdapNameBuilder.newInstance()
+                .add(getBaseDn())
+                .add(getUserSearchBase())
+                .add("uid", accessId)
```
mustaphazorgati, line 202: what happens if the accessId is already a full dn? Do we have a test for this case?

rest/taskana-rest-spring/src/test/java/pro/taskana/common/rest/AccessIdControllerIntTest.java
```diff
@@ -61,6 +61,17 @@ void testQueryGroupsByCn() {
         .containsExactly("cn=ksc-users,cn=groups,OU=Test,O=TASKANA");
   }
 
+  @Test
+  void should_returnEmptyResults_ifInvalidCharacterIsUsedInCondition() {
```
mustaphazorgati, line 65: Sorry :sweat_smile: : I don't understand the test name. 
What in `ksc-teamleads,cn=groups` is invalid?

rest/taskana-rest-spring/src/test/java/pro/taskana/common/rest/AccessIdControllerIntTest.java
```diff
@@ -128,6 +139,23 @@ void should_returnAccessIdsOfGroupsTheAccessIdIsMemberOf_ifAccessIdOfUserIsGiven
                 + "cn=Organisationseinheit KSC,cn=organisation,OU=Test,O=TASKANA");
   }
 
+  @Test
+  void should_returnBadRequest_ifAccessIdOfUserContainsInvalidCharacter() {
```
mustaphazorgati, line 143: Does this test make sense? You want to test something within the LdapClient, but the exception is thrown in the AccessIdController. Thus this test only tests that getGroupsByAccessId is called with a valid accessId

rest/taskana-rest-spring/src/main/java/pro/taskana/common/rest/ldap/LdapClient.java
holgerhagen: It could be either a full dn, then "orFilter.or(new EqualsFilter(getGroupsOfUser(), accessId))" applies, or it is a userid, then this case applies.

rest/taskana-rest-spring/src/test/java/pro/taskana/common/rest/AccessIdControllerIntTest.java
holgerhagen: It is supposed to be a cn and nothing more. LDAP injection aims to prevent, that you can change the search path implictly, as I try in this case.

rest/taskana-rest-spring/src/test/java/pro/taskana/common/rest/AccessIdControllerIntTest.java
holgerhagen: This makes sense in case we do a refactoring. In the end, the test makes sure that this case is correctly handled, regardless the way we do it... 
Ok for you?

rest/taskana-rest-spring/src/main/java/pro/taskana/common/rest/ldap/LdapClient.java
mustaphazorgati: I don't see a matching test for that.

rest/taskana-rest-spring/src/test/java/pro/taskana/common/rest/AccessIdControllerIntTest.java
mustaphazorgati: For now, yes :)

rest/taskana-rest-spring/src/test/java/pro/taskana/common/rest/AccessIdControllerIntTest.java
mustaphazorgati: thanks! :)

rest/taskana-rest-spring/src/main/java/pro/taskana/common/rest/ldap/LdapClient.java
holgerhagen: This test already exists: 
AccessIdControllerIntTest::should_returnAccessIdsOfGroupsTheAccessIdIsMemberOf_ifAccessIdOfGroupIsGiven

mustaphazorgati: ok :) Missed that. Thanks

PR #1162, opened 2020-07-01
TSK-1319: provided google-java-format 1.8.0

mustaphazorgati:
https://taskana.atlassian.net/wiki/spaces/TAS/pages/914522117/Code+Formatting


mustaphazorgati:
@holgerhagen can you please use new jar and test if it still runs with eclipse?

mustaphazorgati:
does not work with jdk8 :(

PR #1173, opened 2020-07-14
TSK-1332: Add HistoryCleanupJob

gitgoodjhe:
Sonarcloud:

 https://sonarcloud.io/dashboard?branch=TSK-1332&id=gitgoodjhe_taskana

NOTE:

Code smell 1: duplicated block readPropertiesFromFile is present in TaskanaEngineConfiguration. The HistoryCleanupJob configuration should not be realized in the engine configuration since not everyone uses the history. Therefore the properties need to be read separately, which produces the code smell. Maybe this is a candidate to refactor into a utility class, since upcoming PRs will also need to read from properties file.

Code smell 2: not gonna fix!

updated documentation:

https://taskana.atlassian.net/wiki/spaces/TAS/pages/687308819/Job+Configuration


lib/taskana-core/src/main/java/pro/taskana/task/internal/jobs/TaskCleanupJob.java
```diff
@@ -98,45 +99,41 @@ public static void initializeSchedule(TaskanaEngine taskanaEngine) {
...
-                .count());
-        countParentTask.merge(task.getParentBusinessProcessId(), 1L, Long::sum);
-      }
 
-      List<String> idsList = new ArrayList<>();
-      numberParentTasksShouldHave.forEach(
-          (k, v) -> {
-            if (v.compareTo(countParentTask.get(k)) == 0) {
-              idsList.add(k);
+      tasksToDelete.forEach(
```
holgerhagen, line 103: Why did you change the TaskCleanupJob?

lib/taskana-core/src/main/java/pro/taskana/common/internal/jobs/AbstractTaskanaJob.java
```diff
@@ -42,6 +45,14 @@ public static TaskanaJob createFromScheduledJob(
         return new TaskCleanupJob(engine, txProvider, job);
       case WORKBASKETCLEANUPJOB:
         return new WorkbasketCleanupJob(engine, txProvider, job);
+      case HISTORYCLEANUPJOB:
```
holgerhagen, line 48: Didn't we agree to change this mechanism to reflection to prevent this kind of hard dependency?

history/taskana-simplehistory-provider/src/main/java/pro/taskana/simplehistory/impl/HistoryCleanupJob.java
```diff
@@ -0,0 +1,337 @@
+package pro.taskana.simplehistory.impl;
```
holgerhagen, line 1: All other jobs reside in a package "jobs". Maybe we should stick to this structure for the history as well?

history/taskana-simplehistory-provider/src/main/java/pro/taskana/simplehistory/impl/HistoryCleanupJob.java
```diff
@@ -0,0 +1,337 @@
...
+    try {
+
+      SimpleHistoryServiceImpl simpleHistoryService =
+          (SimpleHistoryServiceImpl) taskanaHistoryEngine.getTaskanaHistoryService();
+
+      List<HistoryEventImpl> historyEventsToClean =
+          simpleHistoryService
+              .createHistoryQuery()
+              .createdWithin(new TimeInterval(null, createdBefore))
+              .eventTypeIn("TASK_COMPLETED")
```
holgerhagen, line 82: I don't think that this is correct. We do want to delete all events for the tasks and not just the completed event.

holgerhagen (approved):

PR #1175, opened 2020-07-14
TSK-1333: Add history events for cancelled/terminated tasks

gitgoodjhe:
Sonarcloud: https://sonarcloud.io/dashboard?branch=TSK-1333&id=gitgoodjhe_taskana

NOTE:

The test coverage is once again not correct. it should be 100%, since it says the uncovered lines are the constructors of the 2 new events as well as the creation in the terminate/cancel methods of the TaskServiceImpl. These clearly actually get tested in the PR.

### DIFF
--- a/history/taskana-simplehistory-provider/src/test/java/acceptance/events/CreateHistoryEventOnTaskCancellationAccTest.java
+++ b/history/taskana-simplehistory-provider/src/test/java/acceptance/events/CreateHistoryEventOnTaskCancellationAccTest.java
@@ -12,8 +12,6 @@ import org.junit.jupiter.api.extension.ExtendWith;
 import pro.taskana.simplehistory.impl.HistoryEventImpl;
 import pro.taskana.simplehistory.impl.SimpleHistoryServiceImpl;
 import pro.taskana.task.api.TaskService;
-import pro.taskana.task.api.TaskState;
-import pro.taskana.task.api.models.Task;
 
 @ExtendWith(JaasExtension.class)
 class CreateHistoryEventOnTaskCancellationAccTest extends AbstractAccTest {
@@ -23,7 +21,7 @@ class CreateHistoryEventOnTaskCancellationAccTest extends AbstractAccTest {
 
   @Test
   @WithAccessId(user = "admin")
-  void should_CreateCancelledHistoryEvent_When_TaskIsCancelled() throws Exception {
+  void should_CreateCancelledHistoryEvent_When_CancelTaskInStateClaimed() throws Exception {
 
     final String taskId = "TKI:000000000000000000000000000000000001";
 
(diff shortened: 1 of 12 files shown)


history/taskana-simplehistory-provider/src/test/java/acceptance/events/CreateHistoryEventOnTaskCancellationAccTest.java
```diff
@@ -0,0 +1,45 @@
...
+  @WithAccessId(user = "admin")
+  void should_CreateCancelledHistoryEvent_When_TaskIsCancelled() throws Exception {
+
+    final String taskId = "TKI:000000000000000000000000000000000001";
+
+    List<HistoryEventImpl> listEvents = historyService.createHistoryQuery().taskIdIn(taskId).list();
+
+    assertThat(listEvents).isEmpty();
+
+    assertThat(taskService.getTask(taskId).getState()).isEqualTo(TaskState.CLAIMED);
```
holgerhagen, line 34: I don't know what the good practice is. I personally prefer to separate action from assertion to make it more readable. By maybe Sonar complains? 
What do you think?

holgerhagen: BTW: do we need the task to be in state CLAIMED to cancel it?

history/taskana-simplehistory-provider/src/test/java/acceptance/events/CreateHistoryEventOnTaskTerminationAccTest.java
```diff
@@ -0,0 +1,45 @@
...
+  @WithAccessId(user = "admin")
+  void should_CreateTerminatedHistoryEvent_When_TaskIsTerminated() throws Exception {
+
+    final String taskId = "TKI:000000000000000000000000000000000001";
+
+    List<HistoryEventImpl> listEvents = historyService.createHistoryQuery().taskIdIn(taskId).list();
+
+    assertThat(listEvents).isEmpty();
+
+    assertThat(taskService.getTask(taskId).getState()).isEqualTo(TaskState.CLAIMED);
```
holgerhagen, line 34: see above

lib/taskana-core/src/main/java/pro/taskana/spi/history/api/events/task/CancelledEvent.java
```diff
@@ -0,0 +1,13 @@
...
+
+import pro.taskana.task.api.models.Task;
+
+/** Event fired if a task is cancelled. */
+public class CancelledEvent extends TaskEvent {
+
+  public CancelledEvent(String id, Task task, String userId) {
+    super(id, task, userId, null);
+    eventType = "TASK_CANCELLED";
+    created = task.getModified();
```
holgerhagen, line 11: correct would be to use the completed timestamp, which should be identical.

history/taskana-simplehistory-provider/src/test/java/acceptance/events/CreateHistoryEventOnTaskCancellationAccTest.java
gitgoodjhe: Sonar is not complaining. Personally I think as long as the assert fits into 1 LOC the readability should be granted. For me this is similar to returning something from a method. You can instantly return something or declare it as a variable before returning. Sonar tends to not like creating extra variables before returning. What do you think @mustaphazorgati ? 

No the task state can be anything as long as it is not an end state. I could probably remove the check for the task state before terminating the task if you want me to.

lib/taskana-core/src/main/java/pro/taskana/spi/history/api/events/task/CancelledEvent.java
gitgoodjhe: The timestamp is identical, but I will change it to task.getCompleted()

history/taskana-simplehistory-provider/src/test/java/acceptance/events/CreateHistoryEventOnTaskCancellationAccTest.java
holgerhagen: Maybe we should discuss it in the CoP meeting?
Regarding the status: in this case we do not neet the claim and the check for it. But we should have tests for both cancel operations: one from ready state and one from claimed state.

lib/taskana-core/src/main/java/pro/taskana/spi/history/api/events/task/CancelledEvent.java
holgerhagen: You will but you haven't done it yet, correct? Or is this just an displaying issue?

gitgoodjhe: I ave this changed but not pushed yet, since the chances in the asserts are not yet agreed upon. I'm going to push everything together

PR #1179, opened 2020-07-16
TSK-1340: Classification now gets displayed correctly in Workplace

Tristan2357:
Fixed problem, which caused the classification of a task in workplace not to be displayed.

### DIFF
--- a/web/src/app/workplace/models/task.ts
+++ b/web/src/app/workplace/models/task.ts
@@ -1,12 +1,12 @@
-import { Classification } from 'app/shared/models/classification';
 import { Workbasket } from 'app/shared/models/workbasket';
 import { ObjectReference } from './object-reference';
+import { ClassificationSummary } from '../../shared/models/classification-summary';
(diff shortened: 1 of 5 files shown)


web/src/app/workplace/models/task.ts
```diff
@@ -6,7 +6,7 @@ export class Task {
   constructor(public taskId: string,
     public primaryObjRef: ObjectReference = new ObjectReference(),
     public workbasketSummaryResource?: Workbasket,
-    public classificationSummaryResource?: Classification,
+    public classificationSummary?: Classification,
```
sofie29, line 9: Is there a reason why you name the variable classificationSummary instead of classification although the type is only classification?

Tristan2357: The response for a GET of a task returns an object, which has the property classificationSummary, which actually is just a classificationSummary, as i just checked. 
Will update

PR #1182, opened 2020-07-17
TSK-1291: Removed action property in store to make the code cleaner

sofie29:
https://sonarcloud.io/dashboard?branch=TSK-1291&id=sofie29_taskana

### DIFF
--- a/web/src/app/shared/store/classification-store/classification.state.ts
+++ b/web/src/app/shared/store/classification-store/classification.state.ts
@@ -4,18 +4,16 @@ import { take, tap } from 'rxjs/operators';
 import { TaskanaDate } from 'app/shared/util/taskana.date';
 import { CategoriesResponse,
   ClassificationCategoriesService } from '../../services/classification-categories/classification-categories.service';
-import { CreateClassification,
+import { SaveCreatedClassification,
   DeselectClassification,
-  GetClassifications,
+  GetClassifications, CopyClassification, CreateClassification,
   RemoveSelectedClassification,
   RestoreSelectedClassification,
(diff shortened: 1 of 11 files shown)


web/src/app/shared/store/classification-store/classification.state.ts
```diff
@@ -118,33 +113,59 @@ export class ClassificationState implements NgxsAfterBootstrap {
...
+      selectedClassification: initialClassification,
+      badgeMessage: 'Creating new classification'
+    });
+    return of(null);
+  }
+
+  @Action(CopyClassification)
+  newCopyClassification(ctx: StateContext<ClassificationStateModel>): Observable<null> {
+    const copy = { ...ctx.getState().selectedClassification };
+    copy.key = null;
```
Tristan2357, line 163: why key = null, but classificationId = undefined?
i think classificationId should be null too, as we deliberatly set it to nothing

cnguyen-de: other attributes are copied from the selected classification but id and key are not, so that user can write new key value, while id will be generated by backend when submitted. This behavior is similar to workbasket.

Tristan2357: yes, but shouldn't both be set to null, as null indicates the explicit absence of value, while undefined should show, that the value has not been initiated

sofie29: Setting the id to null was also my first intuition but this leads to problems in the tree component since the classificationId is used as idField for the tree nodes.

PR #1186, opened 2020-07-22
Task/1344

mustaphazorgati:
https://sonarcloud.io/dashboard?branch=task%2F1344&id=mustaphazorgati_taskana


lib/taskana-core/src/main/java/pro/taskana/monitor/internal/reports/WorkbasketReportBuilderImpl.java
```diff
@@ -58,6 +64,20 @@ public WorkbasketReport buildReport(TaskTimestamp timestamp)
           monitorQueryItems,
           new DaysToWorkingDaysReportPreProcessor<>(
               this.columnHeaders, converter, this.inWorkingDays));
+
+      Map<String, String> displayMap =
+          taskanaEngine.runAsAdmin(
```
mustaphazorgati, line 69: an alternative to this is to expect that the current user has read access to every workbasket in the report.
A missing access id leads to a missing displayname for that specific workbasket. Is this an assumtion we can make?

PR #1193, opened 2020-07-24
TSK-1290: Research - Refactor MyBatis mapper classes

nkolytschew:
POC
* ClassificationQueryMapperUtil return the query-script as a String


nkolytschew:
For a discussion, if this is too granular, or if we should "generalize" the string more.

lib/taskana-core/src/main/java/pro/taskana/classification/internal/util/ClassificationQueryMapperUtil.java
```diff
@@ -0,0 +1,166 @@
...
+
+  private String getOrderBy() {
+    return "<if test='!orderBy.isEmpty()'>ORDER BY "
+        + "<foreach item='item' collection='orderBy' separator=',' >${item}</foreach></if> ";
+  }
+
+  private StringBuilder getClassificationQueryInConditions() {
+    final StringBuilder query = new StringBuilder(255);
+
+    query
```
holgerhagen, line 65: one positive aspect of the separation of in and like would be to improve readability in this way:

.append("<if test='key != null'        >AND KEY             IN(<foreach item=...</if>")
.append("<if test='parentId != null'>AND PARENT_ID IN(<foreach item=...</if>")

This way you can directly see that it is all the same and if you are missing something. In this case I would be fine with ignoring the limit for the length of the line

lib/taskana-core/src/main/java/pro/taskana/classification/internal/util/ClassificationQueryMapperUtil.java
```diff
@@ -0,0 +1,166 @@
...
+        + CLOSING_WHERE_TAG;
+  }
+
+  private String getOrderBy() {
+    return "<if test='!orderBy.isEmpty()'>ORDER BY "
+        + "<foreach item='item' collection='orderBy' separator=',' >${item}</foreach></if> ";
+  }
+
+  private StringBuilder getClassificationQueryInConditions() {
+    final StringBuilder query = new StringBuilder(255);
```
holgerhagen, line 63: BTW: why do we use String concatination at one place and StringBuffer at another place?

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskQueryImpl.java
```diff
@@ -707,17 +707,17 @@ public TaskQuery orderByClaimed(SortDirection sortDirection) {
   @Override
   public TaskQuery orderByClassificationKey(SortDirection sortDirection) {
     return DB.DB2.dbProductId.equals(getDatabaseId())
-        ? addOrderCriteria("TCLASSIFICATION_KEY", sortDirection)
-        : addOrderCriteria("t.CLASSIFICATION_KEY", sortDirection);
+               ? addOrderCriteria("TCLASSIFICATION_KEY", sortDirection)
```
holgerhagen, line 710: Why do we have to change TaskQueryImpl in this PR?

lib/taskana-core/src/main/java/pro/taskana/classification/internal/util/ClassificationQueryMapperUtil.java
```diff
@@ -0,0 +1,166 @@
+package pro.taskana.classification.internal.util;
+
+public class ClassificationQueryMapperUtil {
```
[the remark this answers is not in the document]
nkolytschew: Using "QueryProvider" as suffix instead. (CRUD Operations might be a thing in the future)

lib/taskana-core/src/main/java/pro/taskana/classification/internal/util/ClassificationQueryMapperUtil.java
```diff
@@ -0,0 +1,166 @@
...
+
+  private String getSelectColumns() {
+    return "SELECT ID, KEY, PARENT_ID, PARENT_KEY, CATEGORY, TYPE, DOMAIN, VALID_IN_DOMAIN, "
+        + "CREATED, NAME, DESCRIPTION, PRIORITY, SERVICE_LEVEL, APPLICATION_ENTRY_POINT, "
+        + "CUSTOM_1, CUSTOM_2, CUSTOM_3, CUSTOM_4, CUSTOM_5, CUSTOM_6, CUSTOM_7, CUSTOM_8";
+  }
+
+  private String getWhereConditions() {
+    return OPENING_WHERE_TAG
+        + getClassificationQueryInConditions()
```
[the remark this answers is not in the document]
nkolytschew: I did this for the show-case to keep the length of the methods "short".
During the discussion we agreed to keep it that way.

lib/taskana-core/src/main/java/pro/taskana/classification/internal/util/ClassificationQueryMapperUtil.java
nkolytschew: It was a one-liner, which was split in 2 lines on "format".
I didn't want to write 
``return new StringBuilder("");``
for just one line.

lib/taskana-core/src/main/java/pro/taskana/classification/internal/util/ClassificationQueryMapperUtil.java
nkolytschew: fixed!
In general I'm not a fan of ignoring rules via annotations and bypassing checkstyle/sonar. But i guess it makes sense for the readability.

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskQueryImpl.java
nkolytschew: We don't. 
Thank you for pointing our. I'm reverting this.

PR #1194, opened 2020-07-25
TSK-1337: removed some code smells

mustaphazorgati:
https://sonarcloud.io/dashboard?branch=code-smells&id=mustaphazorgati_taskana


lib/taskana-core/src/main/java/pro/taskana/TaskanaEngineConfiguration.java
```diff
@@ -232,10 +238,12 @@ public boolean getUseManagedTransactions() {
     return this.useManagedTransactions;
   }
 
+  @SuppressWarnings("unused")
```
gitgoodjhe, line 241: Why do we keep all these unused getters/setters? Sonarcloud ?

PR #1206, opened 2020-07-29
TSK-1338: tasklist now does not refresh after selection of a task

Tristan2357:
https://sonarcloud.io/dashboard?branch=TSK-1338&id=Tristan2357_taskana

### DIFF
--- a/web/src/app/workplace/components/task-master/task-master.component.ts
+++ b/web/src/app/workplace/components/task-master/task-master.component.ts
@@ -39,7 +39,6 @@ export class TaskMasterComponent implements OnInit, OnDestroy {
   });
 
   requestInProgress = false;
-  objectReference: ObjectReference;
   selectedSearchType: Search = Search.byWorkbasket;
 
   destroy$ = new Subject();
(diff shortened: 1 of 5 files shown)


web/src/app/workplace/components/task-master/task-master.component.ts
```diff
@@ -55,19 +55,13 @@ export class TaskMasterComponent implements OnInit, OnDestroy {
...
-          this.selectedId = '';
-        }
-      });
+    this.taskService.taskChangedStream.pipe(takeUntil(this.destroy$)).subscribe((task: Task) => {
+      this.selectedId = task ? task.taskId : '';
+      if (!this.currentBasket) {
+        this.currentBasket = task.workbasketSummary;
+        this.getTasks();
+      }
+    });
```
sofie29, line 64: Why don't we update selectedId and currentBasket like before when taskSelectedStream (getSelectedTask()) changes?

Tristan2357: I used the taskChangedStream, because i assumed it was the same as the selected stream (it worked for my purposes). I changed it to the selectedStream and the list now gets displayed when working on a task.
Good suggestion, thanks

web/src/app/workplace/components/task-master/task-master.component.ts
```diff
@@ -90,9 +84,10 @@ export class TaskMasterComponent implements OnInit, OnDestroy {
 
     this.workplaceService.objectReferenceSelectedStream.pipe(takeUntil(this.destroy$)).subscribe((objectReference) => {
       this.objectReference = objectReference;
-      delete this.currentBasket;
       if (objectReference) {
         this.getTasks();
+      } else {
+        delete this.currentBasket;
```
sofie29, line 90: I thought about this - I think we have to delete the currentBasket even when there is an objectReference. In case we search by type and value, the method getTasks() should be called without a currentBasket. Otherwise the method would filter the tasks by a workbasket (the currentBasket).

web/src/app/workplace/services/task.service.ts
```diff
@@ -91,29 +90,17 @@ export class TaskService {
...
-    if (task.due) {
-      task.due = new Date(task.due).toISOString();
-    }
+  private static convertTasksDatesToGMT(task: Task): Task {
+    const timeAttributes = ['created', 'claimed', 'completed', 'modified', 'planned', 'due'];
+    timeAttributes.forEach((attributeName) => {
+      if (task[attributeName]) {
+        task[attributeName] = new Date(task[attributeName]).toISOString();
+      }
+    });
```
cnguyen-de, line 99: nice

PR #1276, opened 2020-09-22
TSK-1390: Add SPI to allow task processing before creation

gitgoodjhe:
Sonarcloud: https://sonarcloud.io/dashboard?branch=TSK_1390&id=gitgoodjhe_taskana


lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -209,6 +212,10 @@ public Task createTask(Task taskToCreate)
       workbasketService.checkAuthorization(
           task.getWorkbasketSummary().getId(), WorkbasketPermission.APPEND);
 
+      if (CreateTaskPreprocessorManager.isCreateTaskPreprocessorEnabled()) {
```
holgerhagen, line 215: Why did you put it here? I would have expected it even before the opening of the connection to the database...

lib/taskana-core/src/main/java/pro/taskana/spi/taskpreprocessing/api/CreateTaskPreprocessor.java
```diff
@@ -0,0 +1,13 @@
+package pro.taskana.spi.taskpreprocessing.api;
```
holgerhagen, line 1: I would prefer the more simple package pro.taskana.spi.task.
And do we really need ".api"? Isn't SPI already enough?

holgerhagen (changes requested):

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
gitgoodjhe: I wasn't sure if the workbasket/summary would definately never be needed in the SPI method....changed it to before opening the connection of the database now

lib/taskana-core/src/main/java/pro/taskana/spi/taskpreprocessing/api/CreateTaskPreprocessor.java
gitgoodjhe: changed the package to pro.taskana.spi.task. api/internal packages to be consistent with history and task trouting spis. Do you want all of them restructured to without api/internal packages?

holgerhagen (approved):

PR #1284, opened 2020-09-28
TSK-1400: H2 in server mode for rest spring example

gitgoodjhe:


rest/taskana-rest-spring-example-boot/src/main/java/pro/taskana/rest/ExampleRestConfiguration.java
```diff
@@ -31,4 +33,11 @@ public PlatformTransactionManager txManager(DataSource dataSource) {
   public SampleDataGenerator generateSampleData(DataSource dataSource) {
     return new SampleDataGenerator(dataSource, schemaName);
   }
+
+  @Bean(initMethod = "start", destroyMethod = "stop")
```
holgerhagen, line 37: maybe we should make a note that this is only required to let the adapter example connect to the same database? 

And I would you another port as 9090. this is still a rather common port for web applications.

holgerhagen (changes requested):

rest/taskana-rest-spring-example-boot/src/main/java/pro/taskana/rest/ExampleRestConfiguration.java
```diff
@@ -31,4 +33,12 @@ public PlatformTransactionManager txManager(DataSource dataSource) {
   public SampleDataGenerator generateSampleData(DataSource dataSource) {
     return new SampleDataGenerator(dataSource, schemaName);
   }
+
+  // only required to let the adapter example connect to the same database
+  @Bean(initMethod = "start", destroyMethod = "stop")
+  public Server inMemoryH2DatabaseaServer() throws SQLException {
+    return Server.createTcpServer(
+        "-tcp", "-tcpAllowOthers", "-tcpPort", "9095");
```
mustaphazorgati, line 41: should we make this port configurable in our application.properties?

mustaphazorgati (approved):

PR #1286, opened 2020-10-01
TSK-1395: Update Navbar to work without bootstrap 3

Franzi321:


web/src/app/shared/components/nav-bar/nav-bar.component.html
```diff
@@ -1,78 +1,48 @@
-<nav class="navbar main navbar-fixed-top">
-  <div class="navbar  no-border-radius navbar-inverse no-gutter col-xs-12">
-    <div class="pull-left col-sm-3 col-md-4">
-      <button type="button" *ngIf="!showNavbar" class="btn btn-default navbar-toggle show pull-left" (click)="toggleNavBar();"
-        aria-expanded="true" aria-controls="navbar" data-toggle="tooltip" title="Menu">
-        <span class="material-icons md-24 white">menu</span>
+<mat-sidenav-container class="example-container">
```
cnguyen-de, line 1: please use BEM methodology for class name. EG sidenav, child components: sidenav__logout-button, sidenav__user-info

web/src/app/shared/components/nav-bar/nav-bar.component.html
```diff
@@ -1,78 +1,48 @@
...
     <div class="nav-version">
       <p id="taskana-version"> Taskana version: {{version}} </p>
     </div>
-  </div>
-  <div *ngIf="showNavbar" class="backdrop" (click)="toggleNavBar()"></div>
-</nav>
+  </mat-sidenav>
+
+  <mat-sidenav-content>
+    <nav class="navbar no-border-radius navbar-inverse no-gutter col-xs-12">
```
cnguyen-de, line 34: would be cool if you can remove bootstrap classes and replace them with your own css so we don't have to touch navbar again when removing bootstrap completely.

web/src/app/shared/components/user-information/user-information.component.html
```diff
@@ -1,16 +1,14 @@
...
-  </div>
-  <div class="user-info white">
-    <span>Logged as: {{userInformation?.userId}}</span>
-    <button type="button" class="btn btn-default white pull-right transparent" (click)="toggleRoles();" aria-expanded="true" aria-controls="roles">
-      <span>Roles</span>
-    </button>
-  </div>
-  <div class="white pull-right roles col-xs-12" [@toggleDown]="showRoles">
-    <span><i>{{roles}}</i></span>
+<div class="icon">
```
cnguyen-de, line 1: also BEM for class names in this file

cnguyen-de (changes requested):
Please change according to the comments

web/src/app/shared/components/nav-bar/nav-bar.component.html
Franzi321: Shall I replace the whole navbar then? Since <nav> is a bootstrap component. That's also the reason why I left the css.

cnguyen-de: you can use native CSS instead of bootstrap classes is what I mean. <nav> is a native HTML tag, if you mean class name "navbar" then it is a bootstrap class which is basically something like
{
position: fixed;
height: 55px;
width: 100vw; 
}

sofie29:
Class names are written with two underscores according to BEM convention: http://getbem.com/naming/

PR #1498, opened 2021-02-17
TSK-1514: TSK-1515: Fixed and reactivated cypress tests

ptreyer:
---
 
### For the submitter:
- [x] I updated the [documentation](https://taskana.atlassian.net/wiki/spaces/TAS/overview) and will supply links to the specific files
- [ ] I did not update the [documentation](https://taskana.atlassian.net/wiki/spaces/TAS/overview)
- [ ] I included a link to the [SonarCloud branch analysis](https://taskana.atlassian.net/wiki/spaces/TAS/pages/1019969636/SonarCloud+Integration)
- [ ] I added a description of changes on the [current release notes](https://taskana.atlassian.net/wiki/spaces/TAS/pages/1281392672/Current+Release+Notes+Taskana)
- [x] I put the ticket in review

### Verified by the reviewer:
- [x] Commit message format → TSK-XXX: Your commit message.
- [ ] PR fulfills the ticket
- [ ] Readability


web/cypress/integration/classifications/classifications.spec.js
```diff
@@ -1,151 +1,126 @@
 context('TASKANA Classifications', () => {
   beforeEach(() => cy.loginAs('admin'));
 
-  it('should be able to visit Classifications and filter by manual', () => {
+  it('should be possible to edit the Service Level of a classification', () => {
     cy.visit(Cypress.env('appUrl') + Cypress.env('adminUrl') + '/classifications');
+    cy.wait(Cypress.env('pageReload'));
```
mustaphazorgati, line 6: I don't want to blow up the scope of this ticket. But Cypress states that this is a bad practice.
See: https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-Waiting

We should follow those best practices and rework the wait command to use aliases instead of an arbitrary number.

Do you want to rework this in this ticket or create a follow up ticket?

web/cypress/integration/classifications/classifications.spec.js
```diff
@@ -1,151 +1,126 @@
...
-    cy.get('#classification-service-level').clear().type(editedValue);
-
-    cy.get('[title="Save"] > .material-icons').click();
-
-    cy.reload();
-
-    cy.wait(Cypress.env('pageReload'));
-
-    cy.get('#classification-service-level').should('have.value', editedValue);
-  });
```
mustaphazorgati, line 116: Why did you remove this test?

web/cypress/integration/history/history.spec.js
```diff
@@ -1,8 +1,4 @@
 context('TASKANA History', () => {
   beforeEach(() => cy.loginAs('admin'));
 
-  it('should display the history', () => {
-    cy.visit(Cypress.env('appUrl') + '/history');
-    cy.get('.table > form > .table-body > .table-row').should('have.length.greaterThan', 10);
-  });
```
mustaphazorgati, line 7: Why did you remove this test?

web/package.json
```diff
@@ -14,8 +14,9 @@
     "lint:fix": "eslint --ext .ts src --fix",
     "format": "prettier --write \"**/*.{js,ts,css,scss,md,json,yml}\"",
     "format:html": "prettier --write \"**/*.{html,}\"",
-    "e2e": "cypress install && cypress run",
-    "e2e:open": "cypress install && cypress open"
+    "e2e-dev": "cypress install && cypress run",
+    "e2e-dev:open": "cypress install && cypress open",
+    "e2e-standalone": "cypress install && cypress run --config-file cypress-standalone.json"
```
mustaphazorgati, line 19: did you test all three options?

Regarding `e2e-standalone`. Do you think we should start the REST Service from this location here?

mustaphazorgati (changes requested):
Good job! :clap: 

Since I've never worked with Cypress I trust most of your changes and was just able to mark out a couple things :)

mustaphazorgati:
One more things: Please squash your commits down to three. One for fixing the tests, one for the dependency and one for the r rework of the requested changes. Thanks!

web/cypress/integration/classifications/classifications.spec.js
ptreyer: This test still exists and has not been removed.

web/cypress.json
```diff
@@ -1,10 +1,10 @@
 {
   "env": {
-    "baseUrl": "http://localhost:8080/taskana",
```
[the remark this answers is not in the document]
ptreyer: The baseUrl is the url that is called when the login is executed. The appUrl is used for navigation after a successful login.  
So this is not the REST API url. Since we only interact with the frontend in Cypress, there is no direct communication with the REST API url.

If needed, I can rename the baseUrl to loginUrl.

web/package.json
ptreyer: I have successfully tested all three options. 
I think we should not start the REST service from the frontend as we then also need to do a maven build first and also we would not delimit the layers and thus create undesired dependencies.

web/cypress/integration/history/history.spec.js
ptreyer: I have deleted this test as the history plugin is started optionally. But you are right, I should add it again and set a corresponding property in the config files which is checked within the test.

web/cypress/integration/classifications/classifications.spec.js
ptreyer: When using the cy.visit it is not possible to work with an alias. 
However, I have found an alternative solution:
cy.location('hash', {timeout: 10000}).should('include', '/page');

In addition, I can delete the explicit page reloads after saving since this does not correspond to a common user behavior.

web/cypress/integration/classifications/classifications.spec.js
```diff
@@ -1,151 +1,126 @@
...
-        cy.reload();
-        cy.wait(Cypress.env('pageReload'));
-        cy.get('.input-group > .form-control')
-          .invoke('val')
-          .then((newValueOfPriority) => {
-            newValueOfPriority = parseFloat(newValueOfPriority);
-            expect(newValueOfPriority).to.eq(oldPriorityValue + 1);
-          });
-      });
-  });
```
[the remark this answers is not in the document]
ptreyer: I deleted this test because the selection of the 'increase value' field was not possible for me.
If this functionality is not included, the test does not appear to be useful.

web/cypress.json
```diff
@@ -1,15 +1,14 @@
...
-    "monitorUrl": "/monitor",
+    "loginUrl": "http://localhost:4200/taskana",
```
mustaphazorgati, line 3: remove this variable? It is not used for local tests.

mustaphazorgati (approved):

mustaphazorgati (changes requested):
Let's squash the commits :)

mustaphazorgati (approved):

PR #2062, opened 2023-01-17
TSK-1977: fix setting ownerLongName while claiming

ryzheboka:
Sonarcloud: https://sonarcloud.io/summary/new_code?id=ryzheboka_taskana&branch=TSK-1977

---
Release Notes:

```

```
 
### For the submitter:
- [ ] I updated the [documentation](https://taskana.atlassian.net/wiki/spaces/TAS/overview) and will supply links to the specific files
- [x] I did not update the [documentation](https://taskana.atlassian.net/wiki/spaces/TAS/overview)
- [x] I included a link to the [SonarCloud branch analysis](https://taskana.atlassian.net/wiki/spaces/TAS/pages/1019969636/SonarCloud+Integration)
- [ ] After integration of the PR, I added a description of changes on the [current release notes](https://taskana.atlassian.net/wiki/spaces/TAS/pages/1281392672/Current+Release+Notes+Taskana)
- [x] I put the ticket in review

### Verified by the reviewer:
- [ ] Commit message format → TSK-XXX: Your commit message.
- [ ] PR fulfills the ticket
- [ ] Readability

### DIFF
--- a/lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
+++ b/lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
@@ -1107,13 +1107,24 @@ public class TaskServiceImpl implements TaskService {
                 addErrorToBulkLog(this::checkPreconditionsForCompleteTask, bulkLog));
       } else {
         String userId = taskanaEngine.getEngine().getCurrentUserContext().getUserid();
+        String userLongName;
+        if (taskanaEngine.getEngine().getConfiguration().getAddAdditionalUserInfo()) {
+          User user = userMapper.findById(userId);
+          if (user != null) {
+            userLongName = user.getLongName();
+          } else {
+            userLongName = null;
+          }
+        } else {
+          userLongName = null;
+        }
         filteredSummaries =
             filteredSummaries.filter(
                 addErrorToBulkLog(
                     summary -> {
                       if (taskIsNotClaimed(summary)) {
                         checkPreconditionsForClaimTask(summary, true);
-                        claimActionsOnTask(summary, userId, now);
+                        claimActionsOnTask(summary, userId, userLongName, now);
                       }
                     },
                     bulkLog));
@@ -1205,6 +1216,13 @@ public class TaskServiceImpl implements TaskService {
       throws TaskNotFoundException, InvalidStateException, InvalidOwnerException,
           NotAuthorizedException {
     String userId = taskanaEngine.getEngine().getCurrentUserContext().getUserid();
+    String userLongName = null;
+    if (taskanaEngine.getEngine().getConfiguration().getAddAdditionalUserInfo()) {
+      User user = userMapper.findById(userId);
+      if (user != null) {
+        userLongName = user.getLongName();
+      }
+    }
     TaskImpl task;
     try {
       taskanaEngine.openConnection();
@@ -1214,7 +1232,7 @@ public class TaskServiceImpl implements TaskService {
       Instant now = Instant.now();
 
       checkPreconditionsForClaimTask(task, forceClaim);
-      claimActionsOnTask(task, userId, now);
+      claimActionsOnTask(task, userId, userLongName, now);
       taskMapper.update(task);
       if (LOGGER.isDebugEnabled()) {
         LOGGER.debug("Task '{}' claimed by user '{}'.", taskId, userId);
@@ -1335,8 +1353,10 @@ public class TaskServiceImpl implements TaskService {
     return task;
   }
 
-  private static void claimActionsOnTask(TaskSummaryImpl task, String userId, Instant now) {
+  private static void claimActionsOnTask(
+      TaskSummaryImpl task, String userId, String userLongName, Instant now) {
     task.setOwner(userId);
+    task.setOwnerLongName(userLongName);
     task.setModified(now);
     task.setClaimed(now);
     task.setRead(true);
--- a/lib/taskana-core-test/src/test/java/acceptance/task/complete/CompleteTaskAccTest.java
+++ b/lib/taskana-core-test/src/test/java/acceptance/task/complete/CompleteTaskAccTest.java
@@ -19,6 +19,7 @@ import org.junit.jupiter.api.TestTemplate;
 import pro.taskana.classification.api.ClassificationService;
 import pro.taskana.classification.api.models.ClassificationSummary;
 import pro.taskana.common.api.BulkOperationResults;
+import pro.taskana.common.api.TaskanaEngine;
 import pro.taskana.common.api.exceptions.InvalidArgumentException;
 import pro.taskana.common.api.exceptions.TaskanaException;
 import pro.taskana.common.api.security.CurrentUserContext;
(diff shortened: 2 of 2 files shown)


lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -1205,6 +1214,11 @@ private Task claim(String taskId, boolean forceClaim)
       throws TaskNotFoundException, InvalidStateException, InvalidOwnerException,
           NotAuthorizedException {
     String userId = taskanaEngine.getEngine().getCurrentUserContext().getUserid();
+    User user = userMapper.findById(userId);
+    String userLongName = null;
+    if (user != null) {
+      userLongName = userMapper.findById(userId).getLongName();
```
gitgoodjhe, line 1220: why are you making another call to the database here? You already have the user entity from line 1217 to retrieve the long name from, right?

Why is the logic here different than the one above from line 1110-1116? :)

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -1107,13 +1107,22 @@ private BulkOperationResults<String, TaskanaException> completeTasks(
                 addErrorToBulkLog(this::checkPreconditionsForCompleteTask, bulkLog));
       } else {
         String userId = taskanaEngine.getEngine().getCurrentUserContext().getUserid();
+        User user = userMapper.findById(userId);
+        String userLongName;
+        if (user != null) {
+          userLongName = user.getLongName();
+        } else {
+          userLongName = null;
```
gitgoodjhe, line 1115: I think this is unnecessary

gitgoodjhe (changes requested):

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -1107,13 +1107,22 @@ private BulkOperationResults<String, TaskanaException> completeTasks(
                 addErrorToBulkLog(this::checkPreconditionsForCompleteTask, bulkLog));
       } else {
         String userId = taskanaEngine.getEngine().getCurrentUserContext().getUserid();
+        User user = userMapper.findById(userId);
```
gitgoodjhe, line 1110: should we maybe check  if additional User Data is configured?

taskanaEngine.getEngine().getConfiguration().getAddAdditionalUserInfo -> if this is false, then we don't need to make a database call at all, right?

gitgoodjhe (changes requested):

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -1214,7 +1228,8 @@ private Task claim(String taskId, boolean forceClaim)
       Instant now = Instant.now();
 
       checkPreconditionsForClaimTask(task, forceClaim);
-      claimActionsOnTask(task, userId, now);
+      claimActionsOnTask(task, userId, userLongName, now,
+              taskanaEngine.getEngine().getConfiguration().getAddAdditionalUserInfo());
```
gitgoodjhe, line 1232: do we really need that boolean parameter?

gitgoodjhe (changes requested):

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -1335,8 +1350,12 @@ private Task requestChanges(String taskId, boolean force)
     return task;
   }
 
-  private static void claimActionsOnTask(TaskSummaryImpl task, String userId, Instant now) {
+  private static void claimActionsOnTask(TaskSummaryImpl task, String userId, String userLongName,
+                                         Instant now, Boolean additionalUserInfo) {
     task.setOwner(userId);
+    if (additionalUserInfo) {
```
gitgoodjhe, line 1356: is this really necessary?

gitgoodjhe (changes requested):

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
ryzheboka: Should I set ownerLongName to the value of userLongName in any case? It will be just set to null if additionalUserInfo is not set.

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
ryzheboka: No :) Without this parameter, ownerLongName will just be set to null if additionalUserInfo is not set.

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
gitgoodjhe: yes I would prefer that, since it would be null anyways if it is not set at all.

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
ryzheboka: Yes, I will adjust it

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
ryzheboka: If I try to do it differently, the compiler will either say: "Variable 'userLongName' might not have been initialized" or "Variable used in lambda expression should be final or effectively final" when calling claimActionsOnTask later. How can it be avoided?

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
ryzheboka: The additional database call was on accident.  I will fix it :)

The logic is different here, because claimActionsOnTask is not inside a lambda expression. This way, userLongName doesn't have to be final.

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -1205,6 +1216,11 @@ private Task claim(String taskId, boolean forceClaim)
       throws TaskNotFoundException, InvalidStateException, InvalidOwnerException,
           NotAuthorizedException {
     String userId = taskanaEngine.getEngine().getCurrentUserContext().getUserid();
+    User user = userMapper.findById(userId);
```
gitgoodjhe, line 1219: Here we should also check if the additonal user data is activated or not, like in the completeTasks method above

gitgoodjhe (changes requested):

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
ryzheboka: Ok :) I added it

gitgoodjhe (approved):

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
mustaphazorgati: You could initialize userLongName with null in line 1110. Then you don't need the else parts anymore

mustaphazorgati (approved):

arolfes (approved):

Issue #2271, opened 2023-05-25
Commit Hooks won't work without JIRA Tickets


arolfes:
The commit hooks relay on an existing Jira ticket number.
We don't use Jira anymore, so we need to change the commit hooks.

arolfes:
Hello @mustaphazorgati ,

do you have an idea what to do here?

mustaphazorgati:
> Hello @mustaphazorgati ,
> 
> do you have an idea what to do here?

Yes, we have to adjust our commit hooks in https://github.com/Taskana/taskana/tree/master/qa/hooks

PR #2370, opened 2023-08-29
Closes #2366 - add TaskEndstatePreprocessor

### LINKED ISSUE #2366: Add SPI TaskEndstatePreprocessor
Todo:
- define SPI 
- Implement SPI manager
- Integrate SPI into the methods (complete, terminate, cancel)
- Example implementation (setting a variable with the TASKANA endstate that gets interpreted as Camunda variable)
- Testing
- Documentation

ryzheboka:
https://sonarcloud.io/summary/new_code?id=ryzheboka_taskana&branch=TSK-2366

---
Release Notes:

```
Add an SPI that is activated before a Task changes into an endstate. The affected methods are "complete", "cancel" and "terminate"
```
 
### For the submitter:
- [ ] I updated the [documentation](https://taskana.atlassian.net/wiki/spaces/TAS/overview) and will supply links to the specific files
- [x] I did not update the [documentation](https://taskana.atlassian.net/wiki/spaces/TAS/overview)
- [x] I included a link to the [SonarCloud branch analysis](https://taskana.atlassian.net/wiki/spaces/TAS/pages/1019969636/SonarCloud+Integration)
- [ ] After integration of the PR, I added a description of changes on the [current release notes](https://taskana.atlassian.net/wiki/spaces/TAS/pages/1281392672/Current+Release+Notes+Taskana)
- [ ] I put the ticket in review

### Verified by the reviewer:
- [ ] Commit message format → (Closes) #<Issue Number>: Your commit message. i.e. Closes #2271: Your commit message.
- [ ] PR fulfills the ticket
- [ ] Readability

### DIFF
--- a/lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
+++ b/lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
@@ -58,6 +58,7 @@ import pro.taskana.spi.task.internal.BeforeRequestChangesManager;
 import pro.taskana.spi.task.internal.BeforeRequestReviewManager;
 import pro.taskana.spi.task.internal.CreateTaskPreprocessorManager;
 import pro.taskana.spi.task.internal.ReviewRequiredManager;
+import pro.taskana.spi.task.internal.TaskEndstatePreprocessorManager;
 import pro.taskana.task.api.CallbackState;
 import pro.taskana.task.api.TaskCommentQuery;
 import pro.taskana.task.api.TaskCustomField;
@@ -123,6 +124,7 @@ public class TaskServiceImpl implements TaskService {
   private final AfterRequestReviewManager afterRequestReviewManager;
   private final BeforeRequestChangesManager beforeRequestChangesManager;
   private final AfterRequestChangesManager afterRequestChangesManager;
+  private final TaskEndstatePreprocessorManager taskEndstatePreprocessorManager;
 
   public TaskServiceImpl(
       InternalTaskanaEngine taskanaEngine,
@@ -146,6 +148,7 @@ public class TaskServiceImpl implements TaskService {
     this.afterRequestReviewManager = taskanaEngine.getAfterRequestReviewManager();
     this.beforeRequestChangesManager = taskanaEngine.getBeforeRequestChangesManager();
     this.afterRequestChangesManager = taskanaEngine.getAfterRequestChangesManager();
+    this.taskEndstatePreprocessorManager = taskanaEngine.getTaskEndstatePreprocessorManager();
     this.taskTransferrer = new TaskTransferrer(taskanaEngine, taskMapper, this);
     this.taskCommentService =
         new TaskCommentServiceImpl(taskanaEngine, taskCommentMapper, userMapper, this);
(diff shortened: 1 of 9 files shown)


lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -1229,6 +1232,7 @@ private TaskImpl terminateCancelCommonActions(String taskId, TaskState targetSta
     task.setModified(now);
     task.setCompleted(now);
     task.setState(targetState);
+    task = (TaskImpl) taskEndstatePreprocessorManager.processTaskBeforeEndstate(task);
```
gitgoodjhe, line 1235: Please check if the SPI is enabled and only if that's the case execute this line of code

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
```diff
@@ -1552,6 +1556,7 @@ private Task completeTask(String taskId, boolean isForced)
 
       Instant now = Instant.now();
       completeActionsOnTask(task, userId, now);
+      task = (TaskImpl) taskEndstatePreprocessorManager.processTaskBeforeEndstate(task);
```
gitgoodjhe, line 1559: Please check if the SPI is enabled and only if that's the case execute this line of code

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
gitgoodjhe, line 1235: Why did you decide to one time put the SPI processing inside the code that modifies the task (terminateCancelCommonActions) and one time when it doesn't (after completeActionsOnTask)

I understand that otherwise you would have to put the code twice (cancel and terminate), but shouldn't we choose a consistent approach/place? What do you think?

gitgoodjhe (changes requested):

lib/taskana-core/src/main/java/pro/taskana/task/internal/TaskServiceImpl.java
ryzheboka: We do this check inconsistenly for other spis. For example, the line ` task = (TaskImpl) beforeRequestReviewManager.beforeRequestReview(task); ` is executed unconditionally. Same as for TaskEndstatePreprocessor, calling the SPIManager returns the same Task if there is no SPI. That's why it's not necessary to check if the SPI is enabled.  However, some other SPIs are executed conditionally. I think it would be better to make it consistently. Should I change all SPIs that are not checked? Or what would be the best solution?

ryzheboka: In all methods, I put the SPI processing right before the mapper does the updating. The update happens in "terminateCancelCommonActions" but not in  "completeActionsOnTask". That's why the processing is placed differently. We would need to refactor terminateCancelCommonActions or completeActionsOnTask to make it consistent. What do you think about that? :)

gitgoodjhe: I took a closer look at this and it's true. It seems to me, that some SPIs really need the isEnabled-Check and some don't absolutely have to. The ones that don't have to are the ones modifying the task and returning it. The ones that have to are the ones that perform logic additionally to the actual SPI call (performance) e.g. history

However, if we decide to use it the way it is implemented here, do we need the isEnabled method for this SPI at all? The API you mentioned (beforeRequestReview(task)) doesn't have it. All others do and use it either in the code or in the manager itself.

Maybe it's easier to have a short call about this together

gitgoodjhe: Yes we would have to do one of these 2 options, that's what i was trying to imply with a consistent approach. Let's talk about this, I have a favourite but would like to know what you think :)

gitgoodjhe (approved):

gitgoodjhe:
Please remember to add documentation for this new SPI

